PR #1, opened 2020-02-26
Add OAP Server CRDs

hanahmily:
Tips for reviewers, we should pay attention to `oapserver_types.go` because most of the files are generated by kubebuilder.


api/v1alpha1/oapserver_types.go
```diff
@@ -0,0 +1,99 @@
...
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// OAPServerSpec defines the desired state of OAPServer
+type OAPServerSpec struct {
+	// Version of OAP.
+	Version string `json:"version"`
```
withlin: What do you think of using `Template corev1.PodTemplateSpec `json:" template "` `?

hanahmily: It makes much sense to me. If we use PodTemplateSpec, fields following `Config`(NodeSelector, Affinity....) can be merged into it.

api/v1alpha1/oapserver_types.go
```diff
@@ -0,0 +1,86 @@
...
+	// Image is the OAP Server Docker image to deploy.
+	Image string `json:"image,omitempty"`
+	// Count is the number of OAP servers
+	Instances int `json:"instances,imitempty"`
```
wayilau: We need add failedlimits for our  oap instances. then choose the releated policy when  limits reached. I think.

hanahmily: I don't think it's a common option for the user, if they want it, `podTemplate` is a good choice for them to set up `restartPolicy` and `backoffLimit`

PR #6, opened 2020-11-07
Add branch protection and requires checks/reviews before merging

kezhenxu94:


wu-sheng (approved):

kezhenxu94:
Seems the context is wrong

wu-sheng:
> Seems the context is wrong

@kezhenxu94 What is the issue?

kezhenxu94:
> > Seems the context is wrong
> 
> @kezhenxu94 What is the issue?

here is the problem, I fix it in #7 

image

PR #9, opened 2020-11-09
Fix typo in binary name

kezhenxu94:


wu-sheng (approved):
Good. @hanahmily Please re-tag when calling a vote.

PR #12, opened 2020-11-22
Introduce custom metrics apiserver which is an adapter to OAP cluster

hanahmily:
The changes implement [external metrics](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/instrumentation/external-metrics-api.md) to support HPA of Kubernetes:

 * [x] External metrics API implementation
 * [x] Relevant manifests
 * [x] Document
 * [x] Resolve https://github.com/apache/skywalking/issues/5772


doc/custom-metrics-adapter.md
```diff
@@ -0,0 +1,77 @@
...
+    target:
+      ....
+```
+
+ * metric_name: The name of metric generated by OAL or other subsystem.
+ * label: available `label_key`s are `service` and `label`. 
+  
+ The options of label keys are:
+ * `service` varies in different scopes. If the scope is `Service`, `ServiceInstance` and `Endpoint`, the `service` represents
+    `serviceName`, `serviceInstanceName` and `endpointName` respectively.
```
wu-sheng, line 58: How to map to ServiceInstance and EndpointName by a string name? Including how about the relation metrics, such as ServiceRelation.

doc/custom-metrics-adapter.md
```diff
@@ -0,0 +1,77 @@
...
+ * `label` is optional, used for querying multi-labels metrics.
+
+For example if your application is a front gateway, you could add the following section to
```
wu-sheng, line 61:
```suggestion
For example, if your application name is `front_gateway`, you could add the following section to 
```

I think you mean this?

doc/custom-metrics-adapter.md
hanahmily: >How to map to ServiceInstance and EndpointName by a string name? 

How about adding a link to swctl relevant sections in its README? 
 * linear: https://github.com/apache/skywalking-cli#metrics-linear
 * multi-lines : https://github.com/apache/skywalking-cli#metrics-linear, but I use a string `label` instead of a natural number `num` to support meter system.

> Including how about the relation metrics, such as ServiceRelation.

have no plan to implement it. HPA cares about the node metrics instead of the relation ones. Do you have any scenarios for it? feel free to let me know.

wu-sheng: > $ ./bin/swctl --display=graph metrics linear --name=service_instance_resp_time --service "load balancer1.system"

This is how to query `instance` level metrics, which I can't find a way to map your documentation.

> have no plan to implement it. HPA cares about the node metrics instead of the relation ones. Do you have any scenarios for it? feel free to let me know.

Make sense to me.

hanahmily: The adapter is able to opt to the below pieces to query OAP just as the swctl command.
```
- type: External
  external:
    metric:
      name: service_instance_resp_time
      metricSelector:
        matchLabels:
          service: load balancer1.system
          ...
```
the adapter doesn't need a `display` argument, and it could analyze metric type(linear/multiple-lines) from the `listMetric` query of GraphQL.

wu-sheng: https://github.com/apache/skywalking/issues/5890 I think this caused your misunderstanding. Let's hold this PR for a while, we should wait for CLI 0.5.0. Agree?

hanahmily: Noop. swck doesn't depend on a release. once this gets merged, swck could sync with it.

wu-sheng: Then, we need a deeper discussion about what do you plan to distribute. In the current release on the download page, http://skywalking.apache.org/downloads/, there is only a source release, so we always expect users to compile codes by themselves. How would this adaptor be?

hanahmily: > there is only a source release

We release a binary package actually, https://dist.apache.org/repos/dist/release/skywalking/swck/0.1.0/. But it's used to build the release image. The binary can't be run outside kubernetes, that's the reason I didn't add a link to our website. In the next release/0.2.0, I will add adapter binary into it.

The image release strategy is since binary release package contains `operator` and `adapter`, I will add them to a single image called `apache/skywalking-swck`. The default entry point of the image is `operator`, if uses want to run adapter, they should add `command: /bin/adapter` to the pod container node. 

You might notice that there are two different images for the operator and adapter respectively. They are only for dev purposes. The docker file of the release version is going to be submitted soon.

wu-sheng: > We release a binary package actually, https://dist.apache.org/repos/dist/release/skywalking/swck/0.1.0/. But it's used to build the release image.

This is strange, we have binary on the apache mirror system, but no one is able to download it, and no one knows it exists. I think we need to fix this.

Image is another topic, if you want to release that, we need the docker file in the release binary tar ball.

hanahmily:
>
> >  I think we need to fix this.

Yep, will update the site ASAP.

> we need the docker file in the release binary tar ball.


sure, release binary's original duty is to build a convenient image for end
users. Actually, swck's document/guides is based on this image.
 If it doesn't exist, the end-user can't follow these steps indicated on
that.

吴晟 Wu Sheng <notifications@github.com> 于2020年11月25日周三 下午3:42写道：

wu-sheng:
> sure, release binary's original duty is to build a convenient image for end
users. Actually, swck's document/guides is based on this image.
 If it doesn't exist, the end-user can't follow these steps indicated on
that.

Then you need to submit a request to INFRA to set up that first.

hanahmily:
> > sure, release binary's original duty is to build a convenient image for end
> > users. Actually, swck's document/guides is based on this image.
> > If it doesn't exist, the end-user can't follow these steps indicated on
> > that.
> 
> Then you need to submit a request to INFRA to set up that first.

done several weeks ago https://hub.docker.com/r/apache/skywalking-swck

doc/custom-metrics-adapter.md
```diff
@@ -51,12 +51,14 @@ External metrics allow you to autoscale your cluster based on any metric availab
...
   
- The options of label keys are:
- * `service` varies in different scopes. If the scope is `Service`, `ServiceInstance` and `Endpoint`, the `service` represents
-    `serviceName`, `serviceInstanceName` and `endpointName` respectively.
- * `label` is optional, used for querying multi-labels metrics.
+The options of label keys are:
+ * `service` The name of the service.
+ * `instance` The name of the service instance.
+ * `endpoint` The name of the endpoint.
+ * `label` is optional, The labels you need to query, used for querying multi-labels metrics. Unlike [swctl](https://github.com/apache/skywalking-cli#metrics-multiple-linear),
```
wu-sheng, line 60: We could use only one label here, right?

hanahmily: exactly. k8s api doesn't support querying more than one metric, so I only introduce a single label.

wu-sheng (approved):
Looks good. @arugal Would you like to run a test about the HPA?

arugal:
> Looks good. @arugal Would you like to run a test about the HPA?

okay

hanahmily:
@arugal thanks in advance. would you mind submitting your scripts or manifests in testing? That artifacts are valuable when we build e2e next release.

arugal:
> @arugal thanks in advance. would you mind submitting your scripts or manifests in testing? That artifacts are valuable when we build e2e next release.

Okay, I'll submit it

PR #16, opened 2020-12-16
Add unit test case for operator

hanahmily:


.github/workflows/go.yml
```diff
@@ -37,17 +55,30 @@ jobs:
...
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+      - name: Update dependencies
+        run: GOPROXY=https://proxy.golang.org go mod download
+      - name: "install kubebuilder"
+        run: ./hack/install-kubebuilder.sh
+      - name: tests
+        run: make test
   checks:
-    name: build
```
kezhenxu94, line 49: this job is to make the required check status pass, if you want to change the name, may need to also change the one in `.asf.yaml`

hanahmily: done. change the name to `ci` that is more sense to me.

kezhenxu94: The `build` job needs be kept for now to make checks status pass, after this gets merged (`.asf` takes effect`), we can remove it then

kezhenxu94 (approved):

arugal (approved):

PR #21, opened 2021-01-18
Preparation of release 0.2

hanahmily:
These changes intend to prepare to release 0.2:

 - Add Fetcher to fetch Prometheus metrics and push them to OAP server.
 - Some examples to show the usage of Operator.
 - Update CHANGES.md


wu-sheng:
> Add Fetcher to fetch Prometheus metrics and push them to OAP server.

Which metrics are they?

hanahmily:
> > Add Fetcher to fetch Prometheus metrics and push them to OAP server.
> 
> Which metrics are they?

Any metrics from pods with the `prometheus_io_scrape` label.

wu-sheng:
> Any metrics from pods with the prometheus_io_scrape label.

I mean, from SkyWalking perspective, could we handle this?

hanahmily:
> > Any metrics from pods with the prometheus_io_scrape label.
> 
> I mean, from SkyWalking perspective, could we handle this?

Firstly, 

> > Any metrics from pods with the prometheus_io_scrape label.
> 
> I mean, from SkyWalking perspective, could we handle this?

In general, yes. oap has the capacity of handling the metrics matched MAL rules while dropping unknown ones. Users could decide which pods should be scraped by attaching/removing that label. 

The label is from the official example of Prometheus, is well known to users. Istio Prometheus add-on also follows the same way to scrape metrics, users who are familiar with Istio Installation will not be surprised by it. 

While it's a little tricky to hide the label behind Fetcher, maybe we should provide a setting to override the default label key, but Fethcer's role is a helper to guide users on how to transfer Prometheus metrics into OAP. If they have some trouble with Fetcher, the OpenTelemetry operator might be an alternative. I don't intend to add more flexible features to the current Fetcher in pre-1.0 versions.

wu-sheng (approved):

PR #26, opened 2021-03-25
Upgrade skywalking-cli dependency

hanahmily:


pkg/provider/provider.go
```diff
@@ -310,3 +296,24 @@ func (p *externalMetricsProvider) selectGroupResource(namespace string) apischem
 func (p *externalMetricsProvider) getMetricNameWithNamespace(metricName string) string {
 	return strings.Join([]string{p.namespace, metricName}, "|")
 }
+
+// TODO: remove this function once cli move it from internal module to pkg
+func parseScope(entity *swctlapi.Entity) swctlapi.Scope {
```
hanahmily, line 301: @kezhenxu94 this func is copied from cli repo. Could we move it from internal to pkg?

pkg/provider/provider.go
```diff
@@ -310,3 +296,24 @@ func (p *externalMetricsProvider) selectGroupResource(namespace string) apischem
 func (p *externalMetricsProvider) getMetricNameWithNamespace(metricName string) string {
 	return strings.Join([]string{p.namespace, metricName}, "|")
 }
+
+// TODO: remove this function once cli move it from internal module to pkg
```
wu-sheng, line 300: Do we have an issue tracking this?

hanahmily: on it.

pkg/provider/provider.go
kezhenxu94: > @kezhenxu94 this func is copied from cli repo. Could we move it from internal to pkg?

@fgksgf can you take over this? Thanks

fgksgf: @kezhenxu94 No problem, i will do this today

kezhenxu94 (approved):

PR #30, opened 2021-06-16
add feature: java agent injector

dashanji:
[PR_Proposal.pdf](https://github.com/apache/skywalking-swck/files/6663541/PR_Proposal.pdf)


hanahmily:
@dashanji Glade to see this happen so soon. But the process is wrong. 

Before this weekend, I will select one student from the candidates, and didn't promise you will be selected. 

If you are selected, we will discuss the details of this feat. After that, you can move the current PR step. 

I suggest you transfer this to a "draft" before the above steps are done.

wu-sheng:
> Before this weekend, I will select one student from the candidates, and didn't promise you will be selected.

@hanahmily I need to point out, from the foundation and PMC perspective, we don't care about whether this student is selected. Even it isn't, PR in public should be considered as a key point. We are going to accept this PR if it is good enough from a tech perspective.
Notice, we have accepted PMC's PR as it is open in public and good. Read this for the context, https://lists.apache.org/thread.html/r1b84b61347ea87792692fd5653ae23c982283cb4dc1c9c2943b51897%40%3Cdev.skywalking.apache.org%3E.
I have informed the Summer 2021 committee, this kind of thing could happen, please don't do a statement like this. 

In the practice way, you could select the student you prefer, but in the open-source and the Apache way, this PR should not be asked to be drafted, which violate the transparency policy.

hanahmily:
> In the practice way, you could select the student you prefer, but in the open-source and the Apache way, this PR should not be asked to be drafted, which violate the transparency policy.

All right. I would like to withdraw my statement about the candidate's selection. But I need this to be drafted because this PR is not ready to be reviewed without any design document review process. That's from the tech perspective, not my personal preference.

wu-sheng:
> > In the practice way, you could select the student you prefer, but in the open-source and the Apache way, this PR should not be asked to be drafted, which violate the transparency policy.
> 
> All right. I would like to withdraw my statement about the candidate's selection. But I need this to be drafted because this PR is not ready to be reviewed without any design document review process. That's from the tech perspective, not my personal preference.

It is fine from any tech requirement you need as usual.

wu-sheng:
Do we have any progress here?

kezhenxu94:
A head up: this patch seems to assume that Java agent is the only one agent injectable, with https://github.com/apache/skywalking/issues/7381 we will be able to inject Python agent too.

So in this PR the label for pod `swck-agent-injected` should be renamed to `swck-java-agent-injected`

hanahmily:
> Do we have any progress here?

The new design has passed, @dashanji has been writing codes.

hanahmily:
> A head up: this patch seems to assume that Java agent is the only one agent injectable, with [apache/skywalking#7381](https://github.com/apache/skywalking/issues/7381) we will be able to inject Python agent too.
> 
> So in this PR the label for pod `swck-agent-injected` should be renamed to `swck-java-agent-injected`

Updated the title according to this situation

arugal:
image
Use feedback :)

wu-sheng:
OK, after this is ready and merged, I expect the student could write up an English blog about how to use this.

pkg/operator/injector/injector.go
```diff
@@ -0,0 +1,487 @@
...
+	annotation *map[string]string) {
+	// set default value
+	s.needInject = false
+	s.injectContainer = ""
+	s.agentOverlay = false
+
+	// set needInject's value , if the pod has the label "swck-java-agent-injected=true",means need inject
+	if *labels == nil {
+		labels = new(map[string]string)
+		s.needInject = false
```
hanahmily, line 144: Can we return here? looks like we don't need go through the below piece.

pkg/operator/injector/injector.go
```diff
@@ -0,0 +1,487 @@
...
+	}
+
+	switch strings.ToLower((*labels)[labelKeyagentInjector]) {
+	case "true":
```
hanahmily, line 148: Why not a simple `if`? a single branch swich-case is overkill.

pkg/operator/injector/injector.go
```diff
@@ -0,0 +1,487 @@
...
+		for _, c := range containers {
+			if c.Name == v {
+				s.injectContainer = v
+				break
+			}
+		}
+	}
+
+	// set agentOverlay's value
+	switch strings.ToLower((*annotation)[AnnoAgentConfigOverlay.Name]) {
```
hanahmily, line 170: Same as above, swich-case is overkill.

pkg/operator/injector/injector.go
```diff
@@ -0,0 +1,487 @@
...
+		case !s.CreateConfigmap(ctx, r.Client, req.Namespace, &pod.ObjectMeta.Annotations):
+			log.Info("create configmap error!please look the error annotation!")
+		// add jvmStr to EnvVar and do real injection
+		default:
+			if s.jvmAgentConfigStr != "" {
+				s.env.Value = strings.Join([]string{s.env.Value, s.jvmAgentConfigStr}, "=")
+			}
+			s.Inject(pod)
+			log.Info("inject successfully!")
+		}
```
hanahmily, line 465: This piece is hard to understand, I have to read your comment to know the logic, that means ppl can't maintain it.

I suggest using `if-else` or https://en.wikipedia.org/wiki/Chain-of-responsibility_pattern to refactor them.

pkg/operator/injector/injector.go
```diff
@@ -0,0 +1,487 @@
...
+	} else {
+		// if the pod don't have the label "swck-java-agent-injected=true",return ok
+		log.Info("don't inject agent!")
```
hanahmily, line 468: remove `!` from logs.

pkg/operator/injector/injector.go
```diff
@@ -0,0 +1,487 @@
...
+				AnnoConfigmapName.DefaultValue, defaultConfigmapNamespace, err.Error()))
+			return false
+		}
+		// create new configmap and update namespace
+		injectConfigmap := corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      configmapName,
+				Namespace: namespace,
+			},
+			Data: defaultConfigmap.Data,
```
hanahmily, line 410: To validate defaultConfigmap as the annotations.

pkg/operator/injector/injector.go
```diff
@@ -0,0 +1,487 @@
...
+		}
+
+		if err := kubeclient.Create(ctx, &injectConfigmap); err != nil {
+			log.Info(err.Error())
```
hanahmily, line 414: To use `log.Error`

pkg/operator/injector/injector.go
```diff
@@ -0,0 +1,487 @@
...
+	configmap := &corev1.ConfigMap{}
+	configmapName := s.configmapVolume.VolumeSource.ConfigMap.LocalObjectReference.Name
+	// check whether the configmap is existed
+	err := kubeclient.Get(ctx, client.ObjectKey{Namespace: namespace, Name: configmapName}, configmap)
+	// if configmap not exist , get configmap from defaultConfigmapNamespace
+	if err != nil {
+		defaultConfigmap := &corev1.ConfigMap{}
+		if err := kubeclient.Get(ctx, client.ObjectKey{Namespace: defaultConfigmapNamespace,
+			Name: AnnoConfigmapName.DefaultValue}, defaultConfigmap); err != nil {
+			log.Info(err.Error())
```
hanahmily, line 399: To use `log.Error`

pkg/operator/injector/injector.go
```diff
@@ -0,0 +1,487 @@
...
+			config := strings.Join([]string{configName, v}, "=")
+			// add to jvmAgentConfigStr
+			if s.jvmAgentConfigStr != "" {
+				s.jvmAgentConfigStr = strings.Join([]string{s.jvmAgentConfigStr, config}, ",")
+			} else {
+				s.jvmAgentConfigStr = config
+			}
+		}
+	}
+	return true
```
hanahmily, line 383: This func always returns `true`. Please remove the return value.

pkg/operator/injector/injector.go
```diff
@@ -0,0 +1,487 @@
...
+	if pod.Spec.Volumes != nil {
+		pod.Spec.Volumes = append(pod.Spec.Volumes, s.sidecarVolume, s.configmapVolume)
+	} else {
+		pod.Spec.Volumes = []corev1.Volume{s.sidecarVolume, s.configmapVolume}
+	}
+
+	// add volumemount and env to container
+	for i, c := range pod.Spec.Containers {
+		// choose a specific container to inject
+		if s.injectContainer != "" && c.Name == s.injectContainer {
```
hanahmily, line 102: Supposing the first container is not the `injectContainer`, does it get injected? If true, it should be a bug.

pkg/operator/injector/injector.go
```diff
@@ -0,0 +1,487 @@
...
+			} else {
+				pod.Spec.Containers[i].VolumeMounts = []corev1.VolumeMount{s.sidecarVolumeMount,
+					s.configmapVolumeMount}
+			}
+			if c.Env != nil {
+				pod.Spec.Containers[i].Env = append(pod.Spec.Containers[i].Env, s.env)
+			} else {
+				pod.Spec.Containers[i].Env = []corev1.EnvVar{s.env}
+			}
+		}
```
hanahmily, line 129: duplicated lines with the lines in above `if` block

pkg/operator/injector/injector.go
```diff
@@ -0,0 +1,487 @@
...
+func (s *SidecarInjectField) OverlaySidecar(as *Annotations, annotation *map[string]string) bool {
+	var (
+		command       string
+		argOption     string
+		argCommand    string
+		configmapName string
+	)
+
+	switch {
+	case !s.setValue(&s.initcontainer.Name, as, annotation, AnnoInitcontainerName):
```
hanahmily, line 213: Too many branches here. A common way is to store annotations into a `slice`, loop which to validate them.

pkg/operator/injector/injector.go
```diff
@@ -0,0 +1,487 @@
...
+	return ok
+}
+
+// OverlayAgentConfig overlays agent config
+func (s *SidecarInjectField) OverlayAgentConfig(as *Annotations, annotation *map[string]string) bool {
+	// jvmAgentConfigStr init
+	s.jvmAgentConfigStr = ""
+
+	switch {
+	case !s.setJvmAgentStr(as, annotation, AnnoAgentNamespace, false):
```
hanahmily, line 301: The same as above, too cases here.

pkg/operator/injector/injector.go
```diff
@@ -0,0 +1,487 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+const (
+	labelKeyagentInjector       = "swck-java-agent-injected"
+	defaultConfigmapNamespace   = "skywalking-swck-system"
```
hanahmily, line 36: Users can change it when installing the operator. Please add a controller to watch a specific configmap as the default configmap template.

hanahmily (changes requested):
I just take a look at the implementation roughly. The main issue is select container which need to inject and how to get the default configmap.

apis/operator/v1alpha1/oapserver_webhook.go
```diff
@@ -37,6 +37,11 @@ func (r *OAPServer) SetupWebhookWithManager(mgr ctrl.Manager) error {
 		Complete()
 }
 
+// nolint: lll
```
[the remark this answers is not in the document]
dashanji: ok,I will fix it.

cmd/manager/manager.go
```diff
@@ -116,6 +118,12 @@ func main() {
 			setupLog.Error(err, "unable to create webhook", "webhook", "Fetcher")
 			os.Exit(1)
 		}
+		// register a webhook to enable the agent injector，
+		setupLog.Info("registering /mutate-v1-pod webhook")
+		mgr.GetWebhookServer().Register("/mutate-v1-pod",
```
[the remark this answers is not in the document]
dashanji: is "/mutate-operator-skywalking-apache-org-v1alpha1-injector"  ok ?

config/operator/webhook/kustomization.yaml
```diff
@@ -19,6 +19,9 @@ resources:
 - service.yaml
 - manifests.yaml
 
+patchesStrategicMerge:
+- add_nsSelector.yaml
```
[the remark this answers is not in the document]
dashanji: Good idea, I didn't find a suitable name before，and I think "ns_selector_patch.yaml" is fine.

main.go
```diff
@@ -26,10 +26,12 @@ import (
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
```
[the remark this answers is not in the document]
dashanji: Ok, I will do it.

pkg/operator/injector/annotation.go
```diff
@@ -0,0 +1,134 @@
...
+const (
+	namePrefix = "skywalking-swck-"
+)
+
+// nolint
+var (
+	//set sidecar's basic information
+	AnnoInitcontainerName  = setDefaultAnnotation(SidecarInitcontainerName, "inject-sky-agent")
+	AnnoInitcontainerImage = setDefaultAnnotation(SidecarInitcontainerImage,
+		"apache/skywalking-java-agent:8.6.0-jdk8")
```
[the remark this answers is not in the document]
dashanji: Get

pkg/operator/injector/injector.go
dashanji: Yes,it's a bug, if injectContainer is set , the container before injectContainer in Spec will be injected. 
Thanks for correcting.

main.go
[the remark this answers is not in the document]
dashanji: Ok, I think ```kubebuilder create webhook``` is not applicable to our pod injection.Pod is the core type, our webhook should be as shown in the link [webhook-for-core-types](https://cloudnative.to/kubebuilder/reference/webhook-for-core-types.html "webhook-for-core-types").

Therefore, I will create a new ```javaagent_webhook.go``` in the api directory, and put the ```kubebuilder mark```and ```handler function``` into the file, you see if it is feasible.

Please see if it is feasible.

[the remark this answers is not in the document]
hanahmily: It makes senses to me. Go for it

pkg/operator/injector/annotation.go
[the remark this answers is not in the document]
wu-sheng: Would we be able to override this? If so, please document it clearly.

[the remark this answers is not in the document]
dashanji: Yes, user can use their own image. After the PR merger, I will write a detailed document to explain.

pkg/operator/injector/annotation.go
```diff
@@ -0,0 +1,134 @@
...
+	AnnoConfigmapVolumeName           = setDefaultAnnotation(ConfigmapVolumeName, "java-agent-configmap-volume")
+	AnnoConfigmapVolumemountMountpath = setDefaultAnnotation(ConfigmapVolumemountMountpath, "/sky/agent/config")
+	AnnoEnvVarName                    = setDefaultAnnotation(SidecarEnvVarName, "AGENT_OPTS")
+	AnnoEnvVarValue                   = setDefaultAnnotation(SidecarEnvValue,
+		"-javaagent:/sky/agent/skywalking-agent.jar")
+	AnnoInjectContainerName = setDefaultAnnotation(SidecarInjectContainerName, "")
+	//set nil
+	AnnoInjectErrorInfo    = setDefaultAnnotation(SidecarInjectErrorInfo, "")
+	AnnoAgentConfigOverlay = setDefaultAnnotation(SidecarAgentConfigOverlay, "false")
+	//set agent config
```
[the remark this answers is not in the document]
kezhenxu94: @dashanji this list is not complete, please keep an eye on https://github.com/apache/skywalking/pull/7470/files , we are adding the complete config list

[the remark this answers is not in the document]
dashanji: @kezhenxu94 ok, get it, thanks for correction.

pkg/operator/injector/injector.go
dashanji: @hanahmily  Thanks for your valuable suggestions. Under your guidance, I fixed some bugs, refactored the code and added some new features. Please review it when you are free.

dashanji:
The CI error may cause by controller-gen, the reason is the link  [controller-gen error](https://github.com/kubernetes-sigs/controller-runtime/issues/780#issuecomment-713408479). For this special case, can we delete
```$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./apis/..." ``` in ``` Makefile``` and ```CI```?
I think it doesn't affect the use .  Please see if it is feasible.

**Comment by hanahmily** [comment]
> The CI error may cause by controller-gen, the reason is the link [controller-gen error](https://github.com/kubernetes-sigs/controller-runtime/issues/780#issuecomment-713408479). For this special case, can we delete
> `$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./apis/..." ` in ` Makefile` and `CI`?
> I think it doesn't affect the use . Please see if it is feasible.

It will insert the license header to generated APIs. If it causes these failures, we can remove it temporarily. Skywalking-eye will guarantee all headers exist.

**Comment by hanahmily** [comment]
@dashanji What's your plan for this PR, you didn't push anything recently?

**Comment by dashanji** [comment]
> @dashanji What's your plan for this PR, you didn't push anything recently?

Next , I think I will add a CRD and a controller that used to watch the injected pod and then create CR. User can get the final agent configuration through the CR. I'm still writing code currently.

File: pkg/operator/injector/injector.go [hunk 21]
```diff
@@ -293,6 +298,29 @@ func (s *SidecarInjectField) OverlayAgent(a Annotations, ao *AnnotationOverlay,
 	return true
 }
 
+// OverlayOptional overlays optional plugins and move optional plugins to the directory(/plugins)
+// user must ensure that the optional plugins are in the injected container's image
+// Notice , user must specify the optional plugins' version
+// such as spring-cloud-gateway-2.0.x or spring-cloud-gateway-2.1.x
+// the final command will be "mv /optional-plugins/*spring-cloud-gateway-2.0.x*  /plugins/"
```
hanahmily, line 305: Is "cp" here? and Do you check java agent image contains "cp" command? @kezhenxu94 Do we use this command safely in the java agent image?

hanahmily (commented):
Some nits.

pkg/operator/injector/injector.go
kezhenxu94: It's safe to use `cp` command in the Java agent image.

The optional plugins reminds me that we have this config `plugin.exclude_plugins=${SW_EXCLUDE_PLUGINS:}`, and as a Docker images, I'm thinking to move all optional plugins into the default `/plugins` and set `SW_EXCLUDE_PLUGINS` to the optional plugins to disable them, so that we don't need to `cp` / `mv` those files, this is also the original reason why we bring the config

pkg/operator/injector/injector.go
```diff
@@ -293,6 +298,29 @@ func (s *SidecarInjectField) OverlayAgent(a Annotations, ao *AnnotationOverlay,
...
+func (s *SidecarInjectField) OverlayOptional(annotation *map[string]string) {
+	sourcePath := strings.Join([]string{s.SidecarVolumeMount.MountPath, "optional-plugins/"}, "/")
```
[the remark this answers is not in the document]
dashanji: Ok，thanks for reminding

pkg/operator/injector/injector.go
hanahmily: @wu-sheng mentioned that putting all optional plugins into `/plugins` will cause class loading failures. Do we fix it right now?

hanahmily (approved):
LGTM

hanahmily:
@dashanji you need to document how to use this injector in follow-up PRs.

PR #33, opened 2021-09-23
Add java agent injector manual.

dashanji:
This is a manual about how to use the java agent injector.(docs/examples/java-agent-injector.md)


hanahmily (changes requested):
The configmap and annotation template should shrink to a smaller size.

Would you please split the document into two separate files: one, from 1 to 3, goes to `docs/`. another, section 4, keeps in `docs/example`

The first document should contain the following steps:

1. how to install an agent injector
2. how to active agent injection. We have two granularities here: namespace and pod
3. how to config the agent. injector supports a precedence order: annotation, confgmap and default configmap.
4. how to config agent. 
5. how to config sidecar.

config/operator/configmap/configmap.yaml
```diff
@@ -45,17 +45,16 @@ data:
     # The service name in UI
```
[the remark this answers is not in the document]
dashanji: ok , 

> The configmap and annotation template should shrink to a smaller size.
> 
> Would you please split the document into two separate files: one, from 1 to 3, goes to `docs/`. another, section 4, keeps in `docs/example`
> 
> The first document should contain the following steps:
> 
> 1. how to install an agent injector
> 2. how to active agent injection. We have two granularities here: namespace and pod
> 3. how to config the agent. injector supports a precedence order: annotation, confgmap and default configmap.
> 4. how to config agent.
> 5. how to config sidecar.



> The configmap and annotation template should shrink to a smaller size.
> 
> Would you please split the document into two separate files: one, from 1 to 3, goes to `docs/`. another, section 4, keeps in `docs/example`
> 
> The first document should contain the following steps:
> 
> 1. how to install an agent injector
> 2. how to active agent injection. We have two granularities here: namespace and pod
> 3. how to config the agent. injector supports a precedence order: annotation, confgmap and default configmap.
> 4. how to config agent.
> 5. how to config sidecar.

ok , I will clean the code and update doc.

wu-sheng:
@hanahmily Is it better to add menu filr and reorganize this repo docs folder, in order to host docs on the website? More and more repositories have been using this mode.

hanahmily:
Noted with thanks, I am gonna add one

docs/java-agent-injector.md
```diff
@@ -0,0 +1,168 @@
+# Java agent injector Manual
+
+In order to use java agent more natively , we propose the java agent injector to inject the agent sidecar to a pod.
+
+When enabled in a pod's namespace , the injector injects the java agent container at pod creation time using a mutating webhook admission controller. By rendering java agent to a shared volume , containers within the pod can use the java agent.
```
hanahmily, line 5:
```suggestion
When enabled in a pod's namespace or a specific pod, the injector injects the java agent container at pod creation time using a mutating webhook admission controller. By rendering java agent to a shared volume, containers within the pod can use the java agent.
```

docs/java-agent-injector.md
```diff
@@ -0,0 +1,168 @@
...
+
+| Resource  | Label               | Enabled value | Disabled value |
+| --------- | ------------------- | ------------- | -------------- |
+| Namespace | swck-injection      | enabled       | disabled       |
+| Pod       | swck-java-agent-injected | "true"        | "false"        |
+
+The injector is configured with the following logic:
+
+1. If either label is disabled, the pod is not injected.
+2. If two labels are enabled, the pod is injected.
```
hanahmily, line 25: IIRC, either label is enabled, the pod CAN be injected.

The pod granularity only helps users to enable injection for some specific pods which belong to a namespace without the `sw-injection` label.

docs/java-agent-injector.md
```diff
@@ -0,0 +1,168 @@
...
+
+Configmap is described in [kubernetes configmap doc](https://kubernetes.io/docs/concepts/configuration/configmap/).
+
+We need to use configmap to set [agent.config](https://skywalking.apache.org/docs/skywalking-java/latest/en/setup/service-agent/java-agent/configurations/#table-of-agent-configuration-properties), so that we can modify the agent configuration without entering the container. 
+
+If there are different configmap in the namepsace , you can choose a configmap by setting [sidecar annotations](#configure-sidecar); If there is no configmap , the injector will create a default configmap. 
+
+### Default configmap
+
+when we active the injector, it will create a default configmap to overlay the `agent.config` in agent container.
```
hanahmily, line 65:
```suggestion
The injector will create a default configmap to overlay the `agent.config` in the agent container. 
```

docs/java-agent-injector.md
```diff
@@ -0,0 +1,168 @@
...
+| --------- | ------------------- | ------------- | -------------- |
+| Namespace | swck-injection      | enabled       | disabled       |
+| Pod       | swck-java-agent-injected | "true"        | "false"        |
+
+The injector is configured with the following logic:
+
+1. If either label is disabled, the pod is not injected.
+2. If two labels are enabled, the pod is injected.
+
+Follow next steps to active java agent injection.
```
hanahmily, line 27:
```suggestion
Follow the next steps to active java agent injection.
```

docs/java-agent-injector.md
```diff
@@ -0,0 +1,168 @@
...
+
+```shell
+$ kubectl get pod -l swck-java-agent-injected=true
+NAME          READY   STATUS    RESTARTS   AGE
+inject-demo   1/1     Running   0          2d2h
+```
+
+## The ways to configure the agent
+
+The java agent injector support a precedence order to configure the agent:
```
hanahmily, line 45:
```suggestion
The java agent injector supports a precedence order to configure the agent:
```

docs/java-agent-injector.md
```diff
@@ -0,0 +1,168 @@
...
+
+We need to use configmap to set [agent.config](https://skywalking.apache.org/docs/skywalking-java/latest/en/setup/service-agent/java-agent/configurations/#table-of-agent-configuration-properties), so that we can modify the agent configuration without entering the container. 
+
+If there are different configmap in the namepsace , you can choose a configmap by setting [sidecar annotations](#configure-sidecar); If there is no configmap , the injector will create a default configmap. 
+
+### Default configmap
+
+when we active the injector, it will create a default configmap to overlay the `agent.config` in agent container. 
+
+The default configmap is shown as below , one is `agent.service_name` and the string cann't be empty; the other is `collector.backend_service` and it needs to be a legal IP address and port, the other fields need to be guaranteed by users themselves. User can change it as their default configmap.
```
hanahmily, line 67:
```suggestion
The default configmap is shown as below, one is `agent.service_name` and the string can't be empty; the other is `collector.backend_service` and it needs to be a legal IP address and port, the other fields need to be guaranteed by users themselves. Users can change it as their default configmap.
```

docs/java-agent-injector.md
```diff
@@ -0,0 +1,168 @@
...
+    # The service name in UI
+    agent.service_name=${SW_AGENT_NAME:Your_ApplicationName}
+
+    # Backend service addresses.
+    collector.backend_service=${SW_AGENT_COLLECTOR_BACKEND_SERVICES:127.0.0.1:11800}
+
+    # Please refer to https://skywalking.apache.org/docs/skywalking-java/latest/en/setup/service-agent/java-agent/configurations/#table-of-agent-configuration-properties to get more details.
+```
+
+To avoid the default configmap deleting by mistake , we use a configmap controller to watch the default configmap. In addition , if user update it but validate error , such as using wrong backend_service , the controller will create the default configmap.
```
hanahmily, line 81:
```suggestion
To avoid the default configmap deleting by mistake, we use a configmap controller to watch the default configmap. In addition, if the user applies an invalid configuration, such as a malformed `backend_service`, the controller will use the default configmap.
```

docs/examples/java-agent-injector-usage.md
```diff
@@ -0,0 +1,255 @@
+# Java agent injector Usage
+
+In this example , you will learn how to use the java agent injector in three ways.
+
+## Install injector
+
+The java agent injector is builded in the operator , so you need to follow [Operator installation instrument](../../README.md#operator) to install the operator firstly.
```
hanahmily, line 7:
```suggestion
The java agent injector is a component of the operator, so you need to follow [Operator installation instrument](../../README.md#operator) to install the operator firstly.
```

docs/examples/java-agent-injector-usage.md
```diff
@@ -0,0 +1,255 @@
...
+
+## Use default configuration
+
+At first , set the injection label in your namespace as below.
+
+```shell
+kubectl label namespace default(your namespace) swck-injection=enabled
+```
+
+Then add `swck-java-agent-injected: "true"` in the labels of yaml file as below.
```
hanahmily, line 17: Why add this label since we actived injector in the namespace

hanahmily (changes requested):
The issue here is either namespace or pod label can active the injector, rather than both of them.

docs/examples/java-agent-injector-usage.md
kezhenxu94: > Why add this label since we actived injector in the namespace

I think this is needed because only via namespace annotation you don't know which language agent should be injected, you definitely don't want to inject a Java agent to a Python application Pod

hanahmily: if we have multi-language pods in a namespace, the pod granularity is enough. 

The namespace label is "configuration sugar" instead of a constraint. Thanks for your mentioning, the label name attached to a namespace should be `swck-java-injection`.

kezhenxu94: Yes if that's the case we might want to keep only one annotation name, which can be applied in namespace level or Pod level, instead of 2 different annotation name

hanahmily: make sense to me. @dashanji please use the same annotation to both namespace and pod.

dashanji: OK , I hava a small question there, the webhook only support namespaceSelector or objectSelector . If either namespace or pod label can active the injector, we need to get every pod's label or the namespace's label in the webhook, the load of the webhook may increase a lot.

hanahmily: After doing some investigation, the injected pods belong to an intersection of these selectors, I thought it's an union:
https://github.com/kubernetes/apiserver/blob/92392ef22153d75b3645b0ae339f89c12767fb52/pkg/admission/plugin/webhook/generic/webhook.go#L143

All right, let's keep it as is.

docs/java-agent-injector.md
hanahmily: Let's keep it as is.

hanahmily (approved):
LGTM

PR #41, opened 2021-10-31
Fix issues for releasing 0.4.0

hanahmily:
Storage controller:

  * Introduce readiness probe to statefulset
  * Apply authorization to controller
  * Opt to parrelle pod management strategy
  * Add init container to increase "vm.max_map_count"
  * Add password to pks12 key/cert

Java agent injector:

  * Fix backend services validation

UI:

 * Add "http" schema to the backend service URL
cc @dashanji @ESonata

### DIFF
--- a/docs/examples/java-agent-injector-usage.md
+++ b/docs/examples/java-agent-injector-usage.md
@@ -34,7 +34,7 @@ spec:
     spec:
       containers:
       - name: demo1
-        image: dashanji/swck-spring-demo:v0.0.3
+        image: ghcr.io/apache/skywalking-swck-spring-demo:v0.0.1
         command: ["java"]
         args: ["-jar","$(AGENT_OPTS)","-jar","/app.jar"]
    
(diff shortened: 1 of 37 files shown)


docs/examples/java-agent-injector-usage.md
```diff
@@ -34,7 +34,7 @@ spec:
     spec:
       containers:
       - name: demo1
-        image: dashanji/swck-spring-demo:v0.0.3
+        image: ghcr.io/apache/skywalking-swck-spring-demo:v0.0.1
```
hanahmily, line 37: @dashanji I renamed this demo image and push it to GitHub container repository.

BTW, What paths we could access in this app? I got a `404` on the root path.

wu-sheng (approved):

docs/examples/java-agent-injector-usage.md
dashanji: Sorry for not giving instructions，we could use `curl http://localhost:(port-forward IP)/hello` to access it. Next , I will add an e2e test based on `skywalking-infra-e2e`, i think it's more convenient to test.

PR #45, opened 2021-11-08
Fix operator role patch issues

hanahmily:
Using JSON patch instead of patch merge strategy to add a rule which is not supported by kubebuilder


wu-sheng:
Next milestone, 0.5.0?

hanahmily:
> Next milestone, 0.5.0?

yep. i set it up.

wu-sheng (approved):

PR #46, opened 2021-11-09
add e2e test[OAP+UI+Agent]

dashanji:
The PR is to add e2e-test based on the [infra-e2e](https://github.com/apache/skywalking-infra-e2e) so that we can test swck locally.

we can use `make e2e-test` to test the agent and oapserver.


Makefile
```diff
@@ -50,6 +50,13 @@ clean:
 test: generate operator-manifests
 	go test ./... -coverprofile cover.out
 
+# Run e2e-test
+e2e-test:
+	@echo "Prepare e2e..."
+	-hack/prepare-e2e.sh
```
kezhenxu94, line 56: This should be in `e2e.yaml`

hack/prepare-e2e.sh
```diff
@@ -0,0 +1,108 @@
...
+    fi 
+}
+# install kind
+function install_kind()
+{
+    if ! command -v kind &> /dev/null; then
+      echo "please refer https://kind.sigs.k8s.io/docs/user/quick-start to install kind!"
+      $prepare_ok=false
+    fi
+}
```
kezhenxu94, line 50: This is not needed

hack/prepare-e2e.sh
```diff
@@ -0,0 +1,108 @@
...
+
+prepare_ok=true
+# install e2e
+function install_e2e()
+{
+    if ! command -v e2e &> /dev/null; then
+      echo "please refer https://github.com/apache/skywalking-infra-e2e/blob/main/docs/en/contribution/Compiling-Guidance.md to install e2e!"
+      $prepare_ok=false
+    fi
+}
```
kezhenxu94, line 34: This is no needed if https://github.com/apache/skywalking-swck/pull/46/files#r745608218

wankai123 (commented):
It seems not add to Github action workflow?

Makefile
dashanji: ok

hack/prepare-e2e.sh
dashanji: ok, I will delete the function `install_e2e` 、`install_docker ` and `install_kind `, thanks a lot.

dashanji:
> It seems not add to Github action workflow?

Yes, thanks for the reminder, after the storage component test is completed, I will add it together.

wu-sheng:
> > It seems not add to Github action workflow?
> 
> Yes, thanks for the reminder, after the storage component test is completed, I will add it together.

What do you mean by this? Is this PR going to merge or just be WIP?

dashanji:
> > > It seems not add to Github action workflow?
> > 
> > 
> > Yes, thanks for the reminder, after the storage component test is completed, I will add it together.
> 
> What do you mean by this? Is this PR going to merge or just be WIP?

yes, this PR is WIP, the storage component has not been tested yet, I will add it later.

wu-sheng:
You could make a PR in WIP status to avoid this case.

dashanji:
> You could make a PR in WIP status to avoid this case.

Ok, I learned it, thanks a lot.

hanahmily:
@dashanji 

Could you introduce the GitHub action workflow in this? I noticed you mentioned that the `Storage` case should be part of this. But I prefer to split it into two workflows:
 - OAP+UI, which is a quick start for most new users. I would like to introduce it in this.
 - OAP+UI+Storage, which is a comprehensive version. Let's add it in the following-up PRs.

hack/prepare-e2e.sh
```diff
@@ -0,0 +1,82 @@
...
+    if ! command -v kubectl &> /dev/null; then
+      echo "please refer https://kubernetes.io/docs/tasks/tools/#install-kubectl to install kubectl!"
+      $prepare_ok=false
+    fi
+}
+# install swctl
+function install_swctl()
+{
+    if ! command -v swctl &> /dev/null; then
+      echo "please refer https://github.com/apache/skywalking-cli to install swctl!"
```
hanahmily, line 39: We can install it by `go install`, and specify the version as well.

test/e2e/e2e.yaml
```diff
@@ -0,0 +1,106 @@
...
+  method: GET
+
+verify:
+  # verify with retry strategy
+  retry:
+    # max retry count
+    count: 10
+    # the interval between two attempts, e.g. 10s, 1m.
+    interval: 10s
+  cases:
```
hanahmily, line 98: Would you pls add more cases to verify `UI` server? I don't intend to verify how the UI renders the page, but to touch the `ui_host`:`ui_port`/graphql, which would help us know whether the UI server works.

test/e2e/e2e.yaml
```diff
@@ -0,0 +1,106 @@
...
+        kind load docker-image controller:v0.4
+        make operator-install
+        make operator-deploy
+      wait:
+        - namespace: skywalking-swck-system
+          resource: pod
+          for: condition=Ready
+    - name: wait webhook installing
+      command: |
+        sleep 30
```
hanahmily, line 42: How about applying CR in a loop to check webhook is ready? A simple waiting will cause the test to flaky. The Github action task might be very slower than our expectation due to the cloud environment.

test/e2e/e2e.yaml
```diff
@@ -0,0 +1,106 @@
...
+        - namespace: default
+          resource: OAPServer/default
+          for: condition=Available
+        - namespace: default
+          resource: UI/default
+          for: condition=Available
+    - name: setup java agent demo
+      command: |
+        kubectl label namespace default swck-injection=enabled
+        sed 's/oapserver-clusterip/'$(kubectl get svc default-oap -o go-template --template='{{.spec.clusterIP}}')'/' test/e2e/demo.yaml | kubectl create -f -
```
hanahmily, line 56: Please deploy skywalking components to another namespace instead of `default`.

For example, `skywalking-system`. If that, pls to use`skywalking-system-oap.skywalking-system:11800` to set up the backend address of the injected agent.

hanahmily (changes requested):
It's good for the first round. Please fix these items

test/e2e/e2e.yaml
hanahmily: Another path is to drop this step. Add the loop to the next instead. If the `kubectl create` failed, rerun it repeatedly until timeout.

test/e2e/e2e.yaml
dashanji: ok

test/e2e/e2e.yaml
dashanji: ok, I will try it.

test/e2e/e2e.yaml
dashanji: get, thanks.

.github/workflows/go.yml
```diff
@@ -80,9 +80,28 @@ jobs:
         run: ./hack/install-kubebuilder.sh
       - name: tests
         run: make test
+  e2e-tests:
+    name: e2e tests(oap+ui+agent)
+    runs-on: ubuntu-20.04
```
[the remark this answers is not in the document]
dashanji: ok

test/e2e/e2e.yaml
```diff
@@ -0,0 +1,117 @@
...
+
+setup:
+  env: kind
+  file: kind.yaml
+  steps:
+    - name: install cert-manager
+      command: |
+        # kind k8s cluster is in $TMPDIR
+        mkdir -p ~/.kube
+        cp $TMPDIR/e2e-k8s.config ~/.kube/config
```
kezhenxu94, line 24: It's better not to override the users' Kubernetes config file, just set the config context via env var or in command line option

test/e2e/e2e.yaml
```diff
@@ -0,0 +1,117 @@
...
+      expected: verify/metrics.yaml
+    - query: swctl --display yaml --base-url=http://${service_skywalking_system_oap_host}:${service_skywalking_system_oap_12800}/graphql endpoint list --keyword=hello --service-name Your_ApplicationName
+      expected: verify/endpoint.yaml
+    - query: swctl --display yaml --base-url=http://${service_skywalking_system_oap_host}:${service_skywalking_system_oap_12800}/graphql metrics linear --name endpoint_cpm --endpoint-name {GET}/hello --service-name Your_ApplicationName | yq e 'to_entries' -
+      expected: verify/metrics.yaml
+    - query: 'curl -s \
+            --request POST \
+            --header "Content-Type: application/json;charset=UTF-8" \
+            --data ''{"query":"query queryEndpoints($serviceId: ID!, $keyword: String!) {\n    getEndpoints: searchEndpoint(serviceId: $serviceId, keyword: $keyword, limit: 100) {\n      key: id\n      label: name\n    }\n}","variables":{serviceId: "WW91cl9BcHBsaWNhdGlvbk5hbWU=.1", keyword: ""}}'' \
+            http://${service_skywalking_system_ui_host}:${service_skywalking_system_ui_80} 2>&1 | yq e ".data.getEndpoints" -P -'
```
kezhenxu94, line 116: Why do you use a raw `curl` to query, doesn't `swctl endpoint list ......` help?

test/e2e/e2e.yaml
dashanji: ok, get it

.github/workflows/go.yml
```diff
@@ -80,9 +80,28 @@ jobs:
...
+          go-version: 1.17
+        id: go
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+      - name: Update dependencies
+        run: GOPROXY=https://proxy.golang.org go mod download
+      - name: Prepare E2E
+        run: make prepare-e2e-test
+      - name: Run E2E Test
+        uses: apache/skywalking-infra-e2e@main      # always prefer to use a revision instead of `main`.
```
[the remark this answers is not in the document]
dashanji: ok

test/e2e/e2e.yaml
dashanji: I want to simulate the UI request to test whether the UI is normal.

wankai123: you can use `swctl endpoint list` query `${service_skywalking_system_ui_host}:${service_skywalking_system_ui_80}` directly.

dashanji: Make sense to me, thanks a lot.

test/e2e/verify/endpoint.yaml
```diff
@@ -0,0 +1,19 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+{{- contains .}}
+- id: {{ b64enc "Your_ApplicationName" }}.1_{{ b64enc "{GET}/hello" }}
+  name: '{GET}/hello'
```
wankai123, line 18: I‘m curious about the endpoint is `{GET}/hello`, not `GET:/hello`, we have changed spring endpoint name to `METHOD:URI` since java-agent 8.8.0

dashanji: the java agent injected is `java-agent 8.7.0`, should we change it into `java-agent 8.8.0`?

wu-sheng: Yes, it is better to use 8.8.0 agent. The format of endpoint name is not changed randomly, this is a big decision.

dashanji: ok, I will change it.

hack/wait-webhook.sh
```diff
@@ -0,0 +1,41 @@
...
+EOF
+
+timeout $TIMEOUT bash -c -- "\
+    while ! kubectl create -f $MANIFEST 2> /dev/null; \
+    do \
+      sleep 0.1; \
+    done"
+
+# make sure the dummy OAPServer will be deleted
+trap "kubectl delete OAPServer dummy; rm $MANIFEST" 0 2 3 15
```
hanahmily, line 41: It's nice to clean up resources here.

test/e2e/e2e.yaml
```diff
@@ -0,0 +1,117 @@
...
+        mkdir -p ~/.kube
+        cp $TMPDIR/e2e-k8s.config ~/.kube/config
+        kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.3.1/cert-manager.yaml
+      wait:
+        - namespace: cert-manager
+          resource: pod
+          for: condition=Ready
+    - name: install operator
+      command: |
+        export OPERATOR_IMG=controller:v0.4
```
hanahmily, line 32: We don't have to specify the tag in case that this is for the latest code.

test/e2e/kind.yaml
```diff
@@ -0,0 +1,41 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# this config file contains all config fields with comments
+# NOTE: this is not a particularly useful config file
+kind: Cluster
```
hanahmily, line 18: What's the objective for this file?

test/e2e/e2e.yaml
dashanji: ok, get it.

test/e2e/kind.yaml
dashanji: The file is the configuration of the `kind` cluster, we can specify the proper image version of `kind`.

Makefile
hanahmily: @kezhenxu94 Would you pls recheck this?

hanahmily:
@kezhenxu94 @wankai123 @wu-sheng Would you mention reviewing this once more

hanahmily (dismissed):
LGTM

Makefile
kezhenxu94: This is not unused and show be removed

Makefile
```diff
@@ -50,6 +50,18 @@ clean:
 test: generate operator-manifests
 	go test ./... -coverprofile cover.out
 
+# Run e2e-test
+e2e-test:
```
kezhenxu94, line 54: Can you put `hack/prepare-e2e.sh` into `e2e.yaml` `steps` section and remove this target in makefile, so that we can directly issue command `e2e run -c test/e2e/e2e.yaml` instead of wrapping it in `make e2e-test`

kezhenxu94: Take this as an example of how to install required tools in e2e, the idea is not to involve other tools (`make` in this case) as mush as possible

https://github.com/apache/skywalking/blob/b63008c61e1fcb1d69b25a8be7ce99f1d987a175/test/e2e-v2/cases/istio/als/e2e.yaml#L30-L36

kezhenxu94 (commented):
Just one nit, others LGTM, thanks @dashanji

wankai123 (commented):
LGTM

Makefile
dashanji: I have a small question here, I think we need to pull the required images firstly as `(install_images function) in hack/prepare-e2e.sh`, this is why `hack/prepare-e2e.sh` is executed before `e2e run -c test/e2e/e2e.yaml`. For other tools, I think we can add them in the `steps`  section.

kezhenxu94: > I have a small question here, I think we need to pull the required images firstly as `(install_images function) in hack/prepare-e2e.sh`, this is why `hack/prepare-e2e.sh` is executed before `e2e run -c test/e2e/e2e.yaml`. For other tools, I think we can add them in the `steps` section.

```
       - quay.io/jetstack/cert-manager-cainjector:v1.3.1
       - quay.io/jetstack/cert-manager-controller:v1.3.1
       - quay.io/jetstack/cert-manager-webhook:v1.3.1
       - docker.io/kubesphere/kube-rbac-proxy:v0.4.1
       - docker.io/apache/skywalking-oap-server:8.8.1
       - docker.io/apache/skywalking-ui:8.8.1
       - ghcr.io/apache/skywalking-swck-spring-demo:v0.0.1
       - docker.io/apache/skywalking-java-agent:8.8.0-java8
```

These are all public images, right?

For public Docker images, they can be pulled inside KinD when needed, so you don't have to pull firstly

dashanji: Yes, these are public images. But I have tested the `e2e` in the local machine and in the github action. In the local machine, we need to pull the image and load the images to kind, then the `e2e` can work, If we don't do so, the `e2e` may not work. In the github action, it can work whether we load the images to kind. I don’t know if my environment settings cause the `e2e` to fail.

wu-sheng: This looks like some DNS or network issue about docker pull :) I used to face one on local KinD due to accidentally set up proxy for the terminal of running KinD.

hanahmily: >  I don’t know if my environment settings cause the e2e to fail.

I tend to believe that that's your local env issue. Let's ensure e2e is working, then I will help you fix that.

dashanji: OK, thanks a lot.

hanahmily (approved):
LGTM, thank you.

PR #47, opened 2021-11-18
Remove deprecated Kubernetes API dependence

RolandMa1986:
This PR includes the following change:
- Generate CRD with controller-gen 0.7.0 to update CRD API Version to apiextensions.k8s.io/v1
- Change cert-manager API Version cert-manager.io/v1, since the older v1alpha2 is no longer supported by cert-manange 1.x
- Change Ingress API version to networking.k8s.io/v1.

Reffer to apache/skywalking#8136

Signed-off-by: Roland.Ma <rolandma@kubesphere.io>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@
 OPERATOR_IMG ?= controller:latest
 ADAPTER_IMG ?= metrics-adapter:latest
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
-CRD_OPTIONS ?= "crd:trivialVersions=true"
+CRD_OPTIONS ?= crd
 
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
@@ -143,7 +143,7 @@ ifeq (, $(shell which controller-gen))
 	CONTROLLER_GEN_TMP_DIR=$$(mktemp -d) ;\
 	cd $$CONTROLLER_GEN_TMP_DIR ;\
 	go mod init tmp ;\
-	go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.5 ;\
+	go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.7.0 ;\
 	rm -rf $$CONTROLLER_GEN_TMP_DIR ;\
 	}
 CONTROLLER_GEN=$(GOBIN)/controller-gen
--- a/apis/operator/v1alpha1/fetcher_webhook.go
+++ b/apis/operator/v1alpha1/fetcher_webhook.go
@@ -36,7 +36,7 @@ func (r *Fetcher) SetupWebhookWithManager(mgr ctrl.Manager) error {
 }
 
 // nolint: lll
-// +kubebuilder:webhook:path=/mutate-operator-skywalking-apache-org-v1alpha1-fetcher,mutating=true,failurePolicy=fail,groups=operator.skywalking.apache.org,resources=fetchers,verbs=create;update,versions=v1alpha1,name=mfetcher.kb.io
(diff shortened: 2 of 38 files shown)


wu-sheng:
Please notice License and dependency update. The SkyWalking eyes tool is ready for go dependency analysis.
Ping @kezhenxu94 if you face issues when use the tool.

hanahmily:
@RolandMa1986 Thank you for your contribution. As I mentioned, we want to upgrade kubebuilder to v3. I would like to investigate the impact of the API update before reviewing it. Do you have some insights into this task?

wu-sheng:
How about keeping the lowest k8s version in the e2e testing? @hanahmily

hanahmily:
> How about keeping the lowest k8s version in the e2e testing? @hanahmily

the 1.19 used by e2e is the minimal version of most k8s vendor/orchestration support. The could vendor like GKE defaults to 1.21 now.

wu-sheng:
> > How about keeping the lowest k8s version in the e2e testing? @hanahmily
> 
> the 1.19 used by e2e is the minimal version of most k8s vendor/orchestration support. The could vendor like GKE defaults to 1.21 now.

Should we add 1.21 in the e2e too? After all, SWCK is much more k8s sensitive than other components.

hanahmily:
> > > How about keeping the lowest k8s version in the e2e testing? @hanahmily
> > 
> > 
> > the 1.19 used by e2e is the minimal version of most k8s vendor/orchestration support. The could vendor like GKE defaults to 1.21 now.
> 
> Should we add 1.21 in the e2e too? After all, SWCK is much more k8s sensitive than other components.

it makes sense to me. @dashanji is on the last e2e case related to the hpa adapter. Once it's completed, we will assess which versions will be included.

RolandMa1986:
> @RolandMa1986 Thank you for your contribution. As I mentioned, we want to upgrade kubebuilder to v3. I would like to investigate the impact of the API update before reviewing it. Do you have some insights into this task?

@hanahmily I think Upgrade Kubebuilder to v3 doesn't have any conflict with this PR. 

There are 3 parts that will be changed in the Kuberbuilder v3 according to the [upgrading guide](https://book.kubebuilder.io/migration/v2vsv3.html)
1. PROJECT file' layout. I'm sure what was changed exactly.
2. Scaffolded tools. The latest Kubebuilder v3.2.0 still uses `controller-tools 0.7.0` which is the same version used in SWCK. So I think that all the generated manifest will be the same.
3. Go mod dependence. For `controller-runtime` and `k8s.io/api`, I think it depends on which kubebuilder version was used when initializing the PROJECT. But in practice, we just need to consider which Kubernetes version we need to support and choose a compatible version.

wu-sheng:
Notice, with dependency changed, we need to update license file of binary.

RolandMa1986:
> Notice, with dependency changed, we need to update license file of binary.

Is there any document that I can follow up on？ I don't know which file I need to change. The SkyWalking eyes result didn't show any tips:
```
INFO Loading configuration from file: .licenserc.yaml 
INFO No config file is given, using the default config 
INFO Totally checked 476 files, valid: 165, invalid: 0, ignored: 311, fixed: 0 
```

hanahmily:
> Notice, with dependency changed, we need to update license file of binary.

@RolandMa1986 has not changed/updated any dependencies in this. 

And I will hold this PR until the coming metric adapter E2E ensure all components of SWCK with these changes work as expected

wu-sheng:
In the CI process, skywalking-eyes is activated for header checking. At here, https://github.com/apache/skywalking-eyes/tree/v0.2.0#resolve-dependencies-licenses, you could find out how to process license audit(manually).

Eventually, this file should be updated.  https://github.com/apache/skywalking-swck/blob/master/dist/LICENSE#L203 including dependencies with licenses.

hanahmily (approved):
LGTM

PR #48, opened 2021-11-19
fix bug

dashanji:
The PR is to fix two bugs:
1. set the Certificate Signer as `kubernetes.io/legacy-unknown` so that we can use the Kubernetes v1.19+.
2. the Storage Statefulset's  Conditions are always nil as follows. We can compare the Storage Statefulset's ReadyReplicas and Replicas to confirm whether the storage is ready and add the condition manually after that.
image


hanahmily (approved):
LGTM

PR #49, opened 2021-11-21
add e2e-test(oap+ui+internel/external storage)

dashanji:
The PR is to add e2e test about `oap+ui+agent+internel/external storage`.

### DIFF
--- a/test/e2e/deploy-elasticsearch.yaml
+++ b/test/e2e/deploy-elasticsearch.yaml
@@ -1,0 +1,93 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: es
+  labels:
+    k8s-app: es
+spec:
+  replicas: 1
+  serviceName: es
+  selector:
+    matchLabels:
+      k8s-app: es
+  template:
+    metadata:
+      labels:
+        k8s-app: es
+    spec:
+      containers:
+      - name: es
+        image: docker.elastic.co/elasticsearch/elasticsearch:7.5.1
+        env:
+          - name: NAMESPACE
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.namespace
+          - name: POD_NAME
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.name
+        volumeMounts:
+          - name: es-config
+            mountPath: /usr/share/elasticsearch/config/elasticsearch.yml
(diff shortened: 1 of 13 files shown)


test/e2e/deploy-elasticsearch.yaml
```diff
@@ -0,0 +1,93 @@
...
+      k8s-app: es
+  template:
+    metadata:
+      labels:
+        k8s-app: es
+    spec:
+      containers:
+      - name: es
+        image: docker.elastic.co/elasticsearch/elasticsearch:7.5.1
+        env:
```
hanahmily, line 36: You don't have to mount a configmap by setting env var `discovery.type=single-node`

test/e2e/deploy-elasticsearch.yaml
```diff
@@ -0,0 +1,93 @@
...
+  name: es
+data:
+  elasticsearch.yml: |
+    cluster.name: "${NAMESPACE}"
+    node.name: "${POD_NAME}"
+    network.host: 0.0.0.0
+    discovery.seed_hosts: "es-in" 
+    cluster.initial_master_nodes: "es-0"
+---
+apiVersion: v1
```
hanahmily, line 66: This service seems redundant since there is only a single node.

hanahmily (commented):
We could follow an agile way to onboard an ES node

wu-sheng:
A notice, we are using 7.5.1, which is good. We need to avoid newer SSPL and ELv2 release. The only accept way of using 7.10+ and 8.x is test only.

test/e2e/oap-ui-agent-external-storage/e2e.yaml
```diff
@@ -0,0 +1,126 @@
...
+      expected: ../verify/metrics.yaml
+    - query: swctl --display yaml --base-url=http://${service_default_oap_host}:${service_default_oap_12800}/graphql endpoint list --keyword=hello --service-name Your_ApplicationName
+      expected: ../verify/endpoint.yaml
+    - query: swctl --display yaml --base-url=http://${service_default_oap_host}:${service_default_oap_12800}/graphql metrics linear --name endpoint_cpm --endpoint-name GET:/hello --service-name Your_ApplicationName | yq e 'to_entries' -
+      expected: ../verify/metrics.yaml
+    # test ui
+    - query: swctl --display yaml --base-url=http://${service_skywalking_system_ui_host}:${service_skywalking_system_ui_80}/graphql service ls
+      expected: ../verify/service.yaml
+    - query: swctl --display yaml --base-url=http://${service_skywalking_system_ui_host}:${service_skywalking_system_ui_80}/graphql endpoint list --keyword=hello --service-name Your_ApplicationName
+      expected: ../verify/endpoint.yaml
```
wankai123, line 123: you could use `include` to avoid copy the same test cases.
can refer: https://github.com/apache/skywalking/blob/master/test/e2e-v2/cases/simple/jdk/e2e.yaml#L45
or: https://github.com/apache/skywalking/blob/master/test/e2e-v2/cases/kafka/log/e2e.yaml#L45

test/e2e/deploy-elasticsearch.yaml
dashanji: Make sense to me, thanks.

test/e2e/deploy-elasticsearch.yaml
dashanji: yes, I will delete it.

test/e2e/oap-ui-agent-external-storage/e2e.yaml
dashanji: get it, thanks.

wankai123 (commented):
LGTM

wu-sheng:
@wankai123 Please check the `approve` if you want to make this ready to merge. @hanahmily Please recheck.

hanahmily (approved):
LGTM

PR #50, opened 2021-11-24
add adapter e2e

dashanji:
The PR is to add adapter e2e, which means the pod's replicas can be autoscaled according to the data of oap.

### DIFF
--- a/test/e2e/demo.yaml
+++ b/test/e2e/demo.yaml
@@ -45,6 +45,7 @@ spec:
             port: 8085
           initialDelaySeconds: 3
           periodSeconds: 3
+          failureThreshold: 10
 ---
 apiVersion: v1
 kind: Service
--- a/test/e2e/demo.yaml
+++ b/test/e2e/demo.yaml
@@ -39,6 +39,12 @@ spec:
         args: ["$(AGENT_OPTS)","-jar","/app.jar"]
         ports:
           - containerPort: 8085
+        readinessProbe:
+          httpGet:
+            path: /hello
+            port: 8085
+          initialDelaySeconds: 3
+          periodSeconds: 3
 ---
 apiVersion: v1
 kind: Service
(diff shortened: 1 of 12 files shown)


test/e2e/demo.yaml
```diff
@@ -39,6 +39,12 @@ spec:
         args: ["$(AGENT_OPTS)","-jar","/app.jar"]
         ports:
           - containerPort: 8085
+        readinessProbe:
+          httpGet:
+            path: /hello
+            port: 8085
+          initialDelaySeconds: 3
+          periodSeconds: 3
```
hanahmily, line 47: Could you add `failureThreshold: 10` to avoid some potential failures.

test/e2e/oap-agent-adapter-hpa/e2e.yaml
```diff
@@ -0,0 +1,106 @@
...
+        port: 12800
+  timeout: 20m
+
+cleanup:
+  # always never success failure
+  on: always
+
+trigger:
+  action: http
+  interval: 30s
```
hanahmily, line 89: My question is that since the throughput is 2 cpm. Base on 

```
desiredReplicas = ceil[currentReplicas * ( currentMetricValue / desiredMetricValue )]
```

The desriedReplicas should be 1 * (2/5) = 1. The replica number doesn't have a chance to go to 3. Anything I missed here?

hanahmily (changes requested):
The scaling result seems not correct. Would you look it out?

test/e2e/oap-agent-adapter-hpa/e2e.yaml
hanahmily: I speculate that the current metric from oap might not 2. Could you verify it firstly?

test/e2e/demo.yaml
dashanji: get it.

test/e2e/oap-agent-adapter-hpa/e2e.yaml
dashanji: yep, I think the `trigger` part will do 50 times in a very short time, and the interval is between the `trigger` part and the `verify` part. I will add a case to show the cpm(15+), which makes it easier to understand.

hanahmily (approved):
LGTM

PR #52, opened 2021-12-09
Polish documents and introduce webhook health checker

dashanji:
The PR is to do 3 things:
* fix the bug in the configmap controller.
* add webhook healthychecker.
* fix issue in the doc.


wu-sheng:
0.5.0 is on the release vote. @hanahmily Is this good enough for 0.6.0?

operator/main.go
```diff
@@ -164,12 +163,12 @@ func main() {
 		setupLog.Info("/mutate-v1-pod webhook is registered")
 	}
 
-	if err := mgr.AddHealthzCheck("healthz", healthz.Ping); err != nil {
-		setupLog.Error(err, "unable to set up health check")
+	if err := mgr.AddHealthzCheck("webhook", mgr.GetWebhookServer().StartedChecker()); err != nil {
```
hanahmily, line 166: 1. you could not change the checker's name which is referred to at https://github.com/apache/skywalking-swck/blob/master/operator/config/manager/manager.yaml#L58

2. The webhooks are active when `ENABLE_WEBHOOKS` is true. Based on that, you should move the webhook checkers to the above block with other webhooks, and enable `healthz.Ping` when `ENABLE_WEBHOOKS` is false.

hanahmily (changes requested):
Since you introduced the webhook health checker, we had a chance to remove the webhook server validation step in E2E cases.

hanahmily:
> 0.5.0 is on the release vote. @hanahmily Is this good enough for 0.6.0?

yep, we could move this to 0.6.0

wu-sheng:
@hanahmily There is a `v0.4.0-rc0 ` tag found, should you remove it?

hanahmily:
> @hanahmily There is a `v0.4.0-rc0 ` tag found, should you remove it?

pushed by mistake, have deleted it.

operator/main.go
dashanji: Thanks for correcting, updated it. BTW, could we merge the PR into 0.5.0 so that we can use the release version to add the java agent injector in the showcase, as the bug in the configmap controller will affect the injector in the showcase.

hanahmily: all right, I will cancel the 0.5.0 vote.

hanahmily:
@dashanji Could you update `CHANGES.md` as well?

dashanji:
ok, thanks a lot.

dashanji:
updated it.

test/e2e/oap-agent-adapter-hpa/e2e.yaml
```diff
@@ -39,9 +39,6 @@ setup:
         - namespace: skywalking-swck-system
           resource: pod
           for: condition=Ready
-    - name: wait webhook installing
-      command: |
-        bash hack/wait-webhook.sh
```
hanahmily, line 44: Can we remove `wait-webhook.sh`? Is it referred to some place else?

dashanji:
yep, I have removed it in the last commit.

hanahmily (approved):
LGTM

PR #56, opened 2022-01-07
Fix the satellite config not support number error

mrproliu:
When I tried to set the config in the Satellite component, and the config value is a number, then the satellite could not set up.


wu-sheng (approved):

hanahmily (approved):
LGTM

PR #58, opened 2022-01-10
use JAVA_TOOL_OPTIONS to replace AGENT_OPTS

dashanji:
Signed-off-by: dashanji <952129620@qq.com>

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,7 +11,7 @@ Release Notes.
 #### Bugs
 - Update release images to set numeric user id 
 - Fix the satellite config not support number error
-
+- Use env JAVA_TOOL_OPTIONS to replace AGENT_OPTS
 #### Chores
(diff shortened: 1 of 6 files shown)


wu-sheng:
@hanahmily Please confirm whether we still accept PR in 0.6.0 release RC

wu-sheng:
Could you explain or provide some references about differences between `JAVA_TOOL_OPTIONS` and `AGENT_OPTS`

dashanji:
@wu-sheng yep, I think the difference is [here](https://github.com/apache/skywalking-showcase/pull/14#discussion_r775429367).

wu-sheng (approved):

wu-sheng:
Make sense to me. Wait for release manager's confirmation.

hanahmily (approved):
LGTM

PR #62, opened 2022-01-27
Fix java-agent-injector usages link

toffentoffen:
Maybe we can add a mk dead link checker https://github.com/apache/skywalking-java/blob/main/.github/workflows/dead-link-checker.yaml


dashanji (approved):
LGTM

wu-sheng (approved):

PR #63, opened 2022-02-07
add resource limits and requests

dashanji:
The PR is related to https://github.com/apache/skywalking/issues/8258.

### DIFF
--- a/operator/pkg/operator/injector/injector.go
+++ b/operator/pkg/operator/injector/injector.go
@@ -19,6 +19,7 @@ package injector
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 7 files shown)


operator/pkg/operator/injector/injector.go
```diff
@@ -252,10 +258,32 @@ func (s *SidecarInjectField) OverlaySidecar(a Annotations, ao *AnnotationOverlay
...
 
+	// add requests and limits to initcontainer
+	if limitsStr != "nil" {
+		limits := make(corev1.ResourceList)
+		err := json.Unmarshal([]byte(limitsStr), &limits)
+		if err != nil {
+			log.Error(err, "unmarshal limitsStr error")
+			return false
+		}
+		s.Initcontainer.Resources.Limits = limits
```
hanahmily, line 274: Would you please validate the string by a checker?

hanahmily (changes requested):
Could you add the resources request/limit to one of the existing e2e test cases?

operator/pkg/operator/injector/injector.go
dashanji: OK, get it.

hanahmily (approved):
LGTM

PR #65, opened 2022-05-18
Add CRD, Controller, e2e for OAPServer's configuration

dashanji:
Related to https://github.com/apache/skywalking/issues/8921


operator/apis/operator/v1alpha1/oapserverconfig_types.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="Version",type="string",priority=1,JSONPath=".spec.version",description="The version"
+// +kubebuilder:printcolumn:name="Instances",type="string",JSONPath=".status.expectedConfiguredNum",description="The number of expected instance"
+// +kubebuilder:printcolumn:name="Running",type="string",JSONPath=".status.realConfiguredNum",description="The number of running"
+
+// OAPServerConfig is the Schema for the oapserverconfigs API
+type OAPServerConfig struct {
```
hanahmily, line 57: Could you split it into `OAPServerConfig` and `OAPDynamicServerConfig`? It's easy to mess up the controller to put all configurations into an entire and could lead to a too big the yaml to load into the APIServer.

operator/apis/operator/v1alpha1/oapserverconfig_types.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+// OAPServerConfigSpec defines the desired state of OAPServerConfig
+type OAPServerConfigSpec struct {
+	// Version of OAP.
+	//+kubebuilder:validation:Required
+	Version string `json:"version,omitempty"`
+	// StaticConfig holds the OAP server static configuration.
+	// +kubebuilder:validation:Optional
+	StaticConfig []corev1.EnvVar `json:"staticConfig,omitempty"`
```
hanahmily, line 32: `StaticConfig` is more than env vars. Most of the files lay in [resources](https://github.com/apache/skywalking/tree/master/oap-server/server-starter/src/main/resources) are static ones. A potential solution is to put configuration files from the same subdirectory,  or a single in the root directory, to an independent CR.

hanahmily (changes requested):
Your implement tends to put all configurations in one place. But the configuration system of skywalking is too complicated to do this. Please redesign the API to adopt fine-grain CRDs.

dashanji:
> Please update `CHANGES.md` for new feats, dependencies upgrading, etc. You should add examples of setting up configurations.

Copy that.

operator/apis/operator/v1alpha1/oapserverconfig_types.go
dashanji: Indeed, If we want to update the static file, we need to create a CR corresponding to a static file and mount it through Configmap.

operator/apis/operator/v1alpha1/oapserverconfig_types.go
```diff
@@ -0,0 +1,76 @@
...
+	// DynamicConfig holds the OAP server dynamic configuration.
+	// +kubebuilder:validation:Optional
+	DynamicConfig map[string]string `json:"dynamicConfig,omitempty"`
```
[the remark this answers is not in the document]
dashanji: The purpose of my design is to minimize the coupling with the oapserver config. When the oapserver configuration in the main repo changes, there is no need to modify it here, what do you think? BTW，`splitting it into a series` is to deal each [config key](https://skywalking.apache.org/docs/main/latest/en/setup/backend/dynamic-config/#single-configuration) as a field of CR? @hanahmily

operator/apis/operator/v1alpha1/oapserverconfig_types.go
dashanji: Okay.

operator/controllers/operator/oapserverconfig_controller.go
```diff
@@ -0,0 +1,351 @@
...
+		deployment.Spec.Template.Spec.Containers[0].Env = overlay
+		if err := r.Client.Update(ctx, &deployment); err != nil {
+			return labelSelector, fmt.Errorf("failed to update the deployment of OAPServer: %w", err)
+		}
+	}
+
+	// create new static configuration configmap
+	configmap = core.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      oapServer.Name + "-" + oapServerConfig.Spec.Version + "-static-config",
```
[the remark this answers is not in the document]
dashanji: Good idea.

operator/apis/operator/v1alpha1/oapserverconfig_types.go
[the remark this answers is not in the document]
hanahmily: No, you don't have to dive into the details of each file. the CRD should support transforming CRD to the [group config](https://skywalking.apache.org/docs/main/latest/en/setup/backend/dynamic-config-configmap/#group-config), for example:

```yaml
apiVersion: operator.skywalking.apache.org/v1alpha1
kind: OAPServerDynamicConfig
metadata:
  name: core.default.endpoint-name-grouping-openapi
spec:
  customerAPI-v1: value of customerAPI-v1
  productAPI-v1: value of productAPI-v1
```

You could extract the group key from each item by leveraging the CRD to make the whole file neater.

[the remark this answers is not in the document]
dashanji: Make sense to me, in my view, we should support the [group config](https://skywalking.apache.org/docs/main/latest/en/setup/backend/dynamic-config-configmap/#group-config) and the 
 [single config](https://skywalking.apache.org/docs/main/latest/en/setup/backend/dynamic-config/#single-configuration). We can use an `array` to store the data rather than a `map`.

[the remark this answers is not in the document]
hanahmily: Yep, you had better support both of them. Not all config-values are YAML. A single config CRD is good in this case. 

> We should use an array to store the data rather than a map.

Why does it port to an array? Would you elaborate it a bit here?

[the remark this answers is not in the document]
dashanji: Emm, what  I thought before is Using a "config-value" `array` can support more configuration in spec, and we don't need to set the value like `customerAPI-v1` as a field of the CRD. For example, we can support the following CRs.

```yaml
apiVersion: operator.skywalking.apache.org/v1alpha1
kind: OAPServerDynamicConfig
metadata:
  name: core.default.endpoint-name-grouping-openapi.v1
spec:
  - config: core.default.endpoint-name-grouping-openapi.customerAPI-v1
     value: value of customerAPI-v1
  - config: core.default.endpoint-name-grouping-openapi.productAPI-v1
     value: value of productAPI-v1
```

```yaml
apiVersion: operator.skywalking.apache.org/v1alpha1
kind: OAPServerDynamicConfig
metadata:
  name: agent-analyzer.default.slowDBAccessThreshold
spec:
  - config: agent-analyzer.default.slowDBAccessThreshold
     value: default:200,mongodb:50
```

```yaml
apiVersion: operator.skywalking.apache.org/v1alpha1
kind: OAPServerDynamicConfig
metadata:
  name: Skywalking-Dynamic-Config
spec:
  - config: core.default.endpoint-name-grouping-openapi.customerAPI-v1
     value: value of customerAPI-v1
  - config: core.default.endpoint-name-grouping-openapi.productAPI-v1
     value: value of productAPI-v1
  - config: agent-analyzer.default.slowDBAccessThreshold
     value: default:200,mongodb:50
  - config: core.default.apdexThreshold
     value : |-
       default: 500
       # example:
       # the threshold of service "tomcat" is 1s
       # tomcat: 1000
       # the threshold of service "springboot1" is 50ms
       # springboot1: 50
```

So users can put a single configuration in a CR for brevity or put all configurations in a CR for convenience. BTW, 
it seems that we can also use `map`, like the following.

```yaml
apiVersion: operator.skywalking.apache.org/v1alpha1
kind: OAPServerDynamicConfig
metadata:
  name: core.default.endpoint-name-grouping-openapi.v1
spec:
   config:
      core.default.endpoint-name-grouping-openapi.customerAPI-v1: value of customerAPI-v1
      core.default.endpoint-name-grouping-openapi.productAPI-v1: value of productAPI-v1
```

```yaml
apiVersion: operator.skywalking.apache.org/v1alpha1
kind: OAPServerDynamicConfig
metadata:
  name: agent-analyzer.default.slowDBAccessThreshold
spec:
  config:
     agent-analyzer.default.slowDBAccessThreshold: default:200,mongodb:50
```

What do you think about it? Is there a better way or something I missed here?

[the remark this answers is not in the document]
hanahmily: `Array` style is good for me, here is my some suggestions:

- You had better replace `config` with `name`. `config` is the combination of `name` and `value`. 
- As I recommended, you have to remove the global prefix from each item.
- Add root node `data` in the `spec`.

```yaml
apiVersion: operator.skywalking.apache.org/v1alpha1
kind: OAPServerDynamicConfig
metadata:
  name: core.default.endpoint-name-grouping-openapi
spec:
  data:
  - name: customerAPI-v1
     value: value of customerAPI-v1
  - name: productAPI-v1
     value: value of productAPI-v1
```

For those who have only a single item, you could leave an empty `name` as below:

```yaml
apiVersion: operator.skywalking.apache.org/v1alpha1
kind: OAPServerDynamicConfig
metadata:
  name: agent-analyzer.default.slowDBAccessThreshold
spec:
  data:
  -  value: default:200,mongodb:50
```

An all in one version should like 

```yaml
apiVersion: operator.skywalking.apache.org/v1alpha1
kind: OAPServerDynamicConfig
metadata:
  name: Skywalking-Dynamic-Config
spec:
  data:
      - name: core.default.endpoint-name-grouping-openapi.customerAPI-v1
         value: value of customerAPI-v1
      - name: core.default.endpoint-name-grouping-openapi.productAPI-v1
         value: value of productAPI-v1
      - name: agent-analyzer.default.slowDBAccessThreshold
         value: default:200,mongodb:50
      - name: core.default.apdexThreshold
         value : |-
           default: 500
           # example:
           # the threshold of service "tomcat" is 1s
           # tomcat: 1000
           # the threshold of service "springboot1" is 50ms
           # springboot1: 50
````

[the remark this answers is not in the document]
Comment (dashanji) [reply operator/apis/operator/v1alpha1/oapserverconfig_types.go:35]: Thanks for the detailed explain, I understand it. As for static files lay in [resources](https://github.com/apache/skywalking/tree/master/oap-server/server-starter/src/main/resources), I think we can use the same way, the `metadata.name` could be their path and the `spec.data.name` could be the specific file.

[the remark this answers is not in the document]
Comment (lujiajing1126) [reply operator/apis/operator/v1alpha1/oapserverconfig_types.go:35]: It that possible we use [`RawExtension`](https://pkg.go.dev/k8s.io/apimachinery/pkg/runtime#RawExtension) for value?

It is much more flexible. The official `kube-scheduler` uses it for configuration.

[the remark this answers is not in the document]
Comment (dashanji) [reply operator/apis/operator/v1alpha1/oapserverconfig_types.go:35]: Thanks for your suggestion, I find an [example](https://stackoverflow.com/questions/66637260/parse-byte-array-for-yaml-value-kubebuilder) and the code is [here](https://go.dev/play/p/syx8v7gAmDH). 

In my understanding, the [RawExtension](https://kubernetes.io/zh/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#rawextension) can only store the object defined in k8s cluster. 

But the configuration is not defined in k8s cluster. If we use the `RawExtension`, it looks like we need to define an object for each config? I think a string can hold the value simply, what do you think?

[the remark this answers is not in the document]
Comment (lujiajing1126) [reply operator/apis/operator/v1alpha1/oapserverconfig_types.go:35]: > In my understanding, the [RawExtension](https://kubernetes.io/zh/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#rawextension) can only store the object defined in k8s cluster.

No. As long as you mark the field with `// +kubebuilder:pruning:PreserveUnknownFields`, api-server will preserve everything you store in this field.

The `RawExtenstion` is actually serialized to `json` format, tools like `kubectl` will implicitly transform it to `yaml`. So we will see something like,

```yaml
apiVersion: operator.skywalking.apache.org/v1alpha1
kind: OAPServerDynamicConfig
metadata:
  name: Skywalking-Dynamic-Config
spec:
  data:
      - name: core.default.endpoint-name-grouping-openapi.customerAPI-v1
         value: value of customerAPI-v1
      - name: core.default.endpoint-name-grouping-openapi.productAPI-v1
         value: value of productAPI-v1
      - name: agent-analyzer.default.slowDBAccessThreshold
         value: default:200,mongodb:50
      - name: core.default.apdexThreshold
         value:
           default: 500 
```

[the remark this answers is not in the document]
dashanji: @lujiajing1126, make sense to me, I know your meaning. Using `RawExtenstion` can make it easier for users to modify the configuration just like `YAML` rather than `String`. As @hanahmily said, currently, when we create a CR, the fields in it will simply be fulfilled in the configmap.  Later, if we want to verify the configuration, I think it is better to use `RawExtenstion`.  What do you think?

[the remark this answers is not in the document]
lujiajing1126: Since these values are directly mapped to `configmap`s, I think string may be enough now.

hanahmily:
@dashanji, any updates here?

dashanji:
Still coding ^_^

docs/oapserver-configuration.md
```diff
@@ -0,0 +1,284 @@
...
+| CreationTime   | All configurations in one CR, the default value is `false`     |
+| LastUpdateTime | The last time this condition was updated                   |
+
+
+
+## Usage of OAPServerDynamicConfig
+
+There are two ways to configure the OAPServerDynamicConfig. 
+
+- **Global configuration**. Put all the dynamic configurations in one CR, you need to set the field `allInOne` to true.
```
hanahmily, line 134: Can we provide only one mode here? It's not global or single. we could split all configurations into several  CRs. The controller could check CRs don't overlap each other.

dashanji: Do you mean when users create an all-in-one CR, we need to split the CR's data into several CRs that only contain one configuration?

test/e2e/skywalking-components-with-satellite.yaml
```diff
@@ -55,9 +55,9 @@ metadata:
   name: skywalking-system
   namespace: skywalking-system
 spec:
-  version: 8.9.1
+  version: 9.0.0
```
wu-sheng, line 58: 9.0.0 has a critical bug for its internal queue, please use 9.1.0 for testing.

docs/oapserver-configuration.md
```diff
@@ -0,0 +1,279 @@
...
+| File       | The static file in OAP Server, which contains three fields`file.path`、`file.name` and `file.data`.  The `file.path` plus the `file.name`  is the real file that needs to be replaced in the container image, and the `file.data` is the final data in the specific file. |
+
+
+
+## Status of OAPServerConfig 
+
+| Field Name            | Description                                          |
+| --------------------- | ---------------------------------------------------- |
+| ExpectedConfiguredNum | The number of oapserver that need to be configured   |
+| RealConfiguredNum     | The number of oapserver that configured successfully |
```
wu-sheng, line 25: Could you be a little more specific about these two fields? These two are read a little strange. What is `need to be configured`?

docs/oapserver-configuration.md
hanahmily: make sense

docs/oapserver-configuration.md
hanahmily: They represents the replica number set up by the config. It's better to polish the names. How about following `Deployment`'s convention?

* `ExpectedConfiguredNum ` -> `Desired`
* `RealConfiguredNum ` -> `Ready`

wu-sheng: These make more sense.

wu-sheng:
@dashanji Once 0.7.0 is released out, I think we may need a new blog to introduce this as well as https://github.com/apache/skywalking-swck/pull/66?

test/e2e/skywalking-components-with-satellite.yaml
dashanji: Get it.

docs/oapserver-configuration.md
dashanji: Make sense to me, thanks for the correction.

dashanji:
Yeah, a blog does make it clearer to users. BTW, I find some the need to add CRDs and Controllers to SWCK increased, so how about adding a doc to help contributors who want to add CRD and Controller in SWCK focus on the coding, and reduce the time overhead of configuring `kubebuilder`, such as debugging and deployment, etc.

dashanji:
@hanahmily Could you please take a look at this?

docs/oapserver-configuration.md
hanahmily: Please update relevant documents accordingly.

hanahmily (approved):
LGTM

PR #79, opened 2023-01-31
Use  `kubectl apply -k` to install operator and adapter

dashanji:
Kubectl supports applying a kustomization directory from a web page(such as github). It's easier than installing from release binary.


wu-sheng:
You could add docs to reference GitHub, but it should not be the only one. As we have released version on other places.

dashanji:
Updated.

hanahmily (approved):

PR #87, opened 2023-07-09
Deprecate the configmap configurations of the java injector.

dashanji:
Fixes https://github.com/apache/skywalking/issues/10931


hanahmily:
@dashanji We updated OAP. Please fix the conflicts then we will release 0.8.0 once this get merged.

hanahmily (approved):

hanahmily:
@dashanji Would you check out the "internal storage" case in e2e. I rerun it several times, there must be some issues in it.

wu-sheng:
And are the OAP, UI, and database upgraded to latedt by default?

dashanji:
@hanahmily, @wu-sheng. The e2e test can run successfully on my local machine, and I want to debug it on the CI. 

But It looks like we can't add the tmate action as https://github.com/apache/skywalking-swck/actions/runs/5628640535, do you have any idea?

wu-sheng:
@kezhenxu94 any thought?

kezhenxu94:
> @hanahmily, @wu-sheng. The e2e test can run successfully on my local machine, and I want to debug it on the CI.
> 
> But It looks like we can't add the tmate action as https://github.com/apache/skywalking-swck/actions/runs/5628640535, do you have any idea?

Don’t over use those tools, collect logs and look into the logs first. It’s time/resource consuming to use that tool to log into the CI machine. now the CI failed and the logs show Pod is not ready in time. next is to collect the pod logs and see whether there is error in it, this is not only useful for this time but also for future

kezhenxu94:
FYI about how to collect logs https://github.com/apache/skywalking/blob/e0068154c1253500720651dca5c80c3ae10ca87b/.github/workflows/skywalking.yaml#L687

dashanji:
> FYI about how to collect logs https://github.com/apache/skywalking/blob/e0068154c1253500720651dca5c80c3ae10ca87b/.github/workflows/skywalking.yaml#L687

Make sense to me. Thanks a lot!

hanahmily (approved):

PR #90, opened 2023-08-28
Fix number env value format error in template yaml.

weixiang1862:


dashanji (approved):
LGTM

wu-sheng:
I am just trying to rerun CIs, two cases are falling.

weixiang1862:
There is still one case failed, it passed on my local.

PR #93, opened 2023-08-31
Add startupProbe, livenessProbe and readinessProbe for OAPServer to fix e2e errors

dashanji:
As titled.

### DIFF
--- a/operator/pkg/operator/manifests/oapserver/templates/deployment.yaml
+++ b/operator/pkg/operator/manifests/oapserver/templates/deployment.yaml
@@ -25,7 +25,6 @@ metadata:
     operator.skywalking.apache.org/oap-server-name: {{ .Name }}
     operator.skywalking.apache.org/application: oapserver
     operator.skywalking.apache.org/component: deployment
-
 spec:
   replicas: {{ .Spec.Instances }}
   minReadySeconds: 5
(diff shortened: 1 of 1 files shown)


.github/workflows/go.yml
```diff
@@ -75,60 +75,68 @@ jobs:
...
       - name: Install Go
         uses: actions/setup-go@v3
         with:
           go-version: '1.20'
         id: go
-      - name: Check out code into the Go module directory
-        uses: actions/checkout@v3
       - name: Run E2E Test(oap+ui+agent)
-        uses: apache/skywalking-infra-e2e@996ed8902e941e2883fcf0ac5b3090364942d205      # always prefer to use a revision instead of `main`.
+        uses: apache/skywalking-infra-e2e@main      # always prefer to use a revision instead of `main`.
```
wu-sheng, line 88: We should lock the version. You could bump up the commit ID, rather than the `main` branch.

.github/free-disk-space.sh
```diff
@@ -0,0 +1,51 @@
...
+sudo apt-get clean
+df -h
+echo "Removing large directories"
+# deleting 15GB
+rm -rf /usr/share/dotnet/
+sudo rm -rf /usr/share/dotnet
+sudo rm -rf /usr/local/lib/android
+sudo rm -rf /opt/ghc
+sudo rm -rf /opt/hostedtoolcache/CodeQL
+df -h
```
wu-sheng, line 51: Could you reduce the log output, rather than deleting files?

.github/workflows/go.yml
dashanji: Make sense to me.

.github/free-disk-space.sh
dashanji: Thanks for the advice, maybe we could reduce the log first, I will try it later.

dashanji:
After debugging for a while, I think the e2e test failed as the OAPServer is not ready for port 12800.

wu-sheng:
So, it is healthy status check issue?

dashanji:
> So, it is healthy status check issue?

Yes. So far it looks like this.

wu-sheng (approved):

PR #95, opened 2023-10-11
Use append rather than replace policy for JAVA_TOOL_OPTIONS

robberphex:
When there is already `JAVA_TOOL_OPTIONS`, i.e. `JAVA_TOOL_OPTIONS=-javaagent:/tmp/a.jar`, currently skywalking-swck will replace it to `JAVA_TOOL_OPTIONS=-javaagent:/path/skywalking.jar`, which override user's setting.

skywalking should append to JAVA_TOOL_OPTIONS, like `JAVA_TOOL_OPTIONS=-javaagent:/tmp/a.jar -javaagent:/path/skywalking.jar`.

### DIFF
--- a/operator/apis/operator/v1alpha1/swagent_webhook.go
+++ b/operator/apis/operator/v1alpha1/swagent_webhook.go
@@ -107,7 +107,7 @@ func (r *SwAgent) setDefault() {
 			r.Spec.JavaSidecar.Args = append(r.Spec.JavaSidecar.Args, "-c")
 			r.Spec.JavaSidecar.Args = append(r.Spec.JavaSidecar.Args, "mkdir -p /sky/agent && cp -r /skywalking/agent/* /sky/agent")
 		}
-		r.setOrAddEnv("JAVA_TOOL_OPTIONS", " -javaagent:/sky/agent/skywalking-agent.jar")
+		r.setOrAppendEnv("JAVA_TOOL_OPTIONS", " -javaagent:/sky/agent/skywalking-agent.jar")
 
 		// default values for shared volume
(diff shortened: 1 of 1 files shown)


wu-sheng:
Please explain your use cases. I am not following.

operator/apis/operator/v1alpha1/swagent_webhook.go
```diff
@@ -127,10 +127,10 @@ func (r *SwAgent) setOrAddEnv(envKey string, envValue string) {
 	}
 }
 
-func (r *SwAgent) setEnvIfExists(envs *[]corev1.EnvVar, envKey string, envValue string) bool {
+func (r *SwAgent) appendEnvIfExists(envs *[]corev1.EnvVar, envKey string, envValue string) bool {
 	for _, env := range *envs {
 		if strings.EqualFold(env.Name, envKey) {
-			env.Value = envValue
+			env.Value += envValue
```
dashanji, line 133: I think it's not a good idea to just append the string here.

dashanji:
Hi @robberphex  What do you think?

wu-sheng:
@robberphex Are you still working on this?

robberphex:
Sorry for late reply, updated description, @wu-sheng

wu-sheng:
@robberphex I can't see new commits submitted. What do you mean `updated`?

dashanji:
> When there is already `JAVA_TOOL_OPTIONS`.

The current design is to configure the agent via env rather than `JAVA_TOOL_OPTIONS`.

@Duncan-tree-zhou What do you think?

dashanji (approved):

PR #100, opened 2023-11-05
Add a getting started doc for swck

dashanji:
Add a document to help users quickly get started with swck, including the following aspects:
- Create a local kubernetes cluster.
- Build the docker image of operator and deploy it on the kubernetes cluster.
- Deploy the skywalking components: OAPServer and UI.
- Use the injector to inject the java agent to the application pod.
- Access the UI and get the metric info.

PR #105, opened 2023-12-01
Overlay javaagent by pod annotation.

weixiang1862:
I use workload annotations to inject agent agent name and collector backend_service, I fount javaagent can not get the correct value:
```
[root@15 ~]# kubectl get javaagent -A
NAMESPACE     NAME                     PODSELECTOR    SERVICENAME            BACKENDSERVICE
gray-demo     app-consumer-javaagent   app=consumer   Your_ApplicationName   127.0.0.1:11800
``` 

So I add a method `injectConfigByPodAnnotations` after `injectConfigBySwAgent`, then javaagent_controller can overlay config by using pod annotations:
```
[root@15 ~]# kubectl get javaagent -A
NAMESPACE     NAME                     PODSELECTOR    SERVICENAME                   BACKENDSERVICE
gray-demo     app-consumer-javaagent   app=consumer   cluster01-default::consumer   192.168.6.71:11800
```

### DIFF
--- a/operator/controllers/operator/javaagent_controller.go
+++ b/operator/controllers/operator/javaagent_controller.go
@@ -98,6 +98,7 @@ func (r *JavaAgentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	}
 	config := map[string]string{}
 	r.injectConfigBySwAgent(lastMatchedSwAgent, config)
+	r.injectConfigByPodAnnotations(pod.ObjectMeta.Annotations, config)
 
 	// only get the first selector label from labels as podselector
 	labels := pod.Labels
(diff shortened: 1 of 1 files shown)


dashanji:
Hi, @weixiang1862. Thanks for the pr.

It's a bug that the javaagent can not get the correct value. But I suggest changing the code as shown https://github.com/apache/skywalking-swck/pull/100/files#diff-a2f954e637bdd44c3be61ae1617b036e80fd238d321591e3cf704fb292418cdcR101

Thus, you can get all configurations of pod annotations in the java agent rather than the two configurations `backend_service` and `service_name`.

weixiang1862:
> Hi, @weixiang1862. Thanks for the pr.
> 
> It's a bug that the javaagent can not get the correct value. But I suggest changing the code as shown https://github.com/apache/skywalking-swck/pull/100/files#diff-a2f954e637bdd44c3be61ae1617b036e80fd238d321591e3cf704fb292418cdcR101
> 
> Thus, you can get all configurations of pod annotations in the java agent rather than the two configurations `backend_service` and `service_name`.

Thanks, your code implementation is better. 
I did not notice you hava already fix it, I will close this pr.

dashanji:
> Thanks, your code implementation is better. I did not notice you hava already fix it, I will close this pr.

Hi, @weixiang1862 please don't worry about this. I mean you can reuse the code implementation here.

BTW, the observability of swagent is not enough, it only supports the `SERVICENAME ` and `BACKENDSERVICE ` rather than all configurations of java agent. Do you have any interest in it?

weixiang1862:
> > Thanks, your code implementation is better. I did not notice you hava already fix it, I will close this pr.
> 
> Hi, @weixiang1862 please don't worry about this. I mean you can reuse the code implementation here.
> 
> BTW, the observability of swagent is not enough, it only supports the `SERVICENAME ` and `BACKENDSERVICE ` rather than all configurations of java agent. Do you have any interest in it?

Yes, I will do some research first and share my ideas here.

PR #106, opened 2023-12-05
Remove SwAgent default env JAVA_TOOL_OPTIONS.

weixiang1862:
From the [swck doc](https://skywalking.apache.org/docs/skywalking-swck/next/java-agent-injector/#configure-sidecar), `JAVA_TOOL_OPTIONS` env name can be overlayed by annotation `sidecar.skywalking.apache.org/env.Name`. 

In SwAgent webhook default func, env field will be innitialized with `JAVA_TOOL_OPTIONS`, then it will be assign to sidecar_inject_field envs.

If I overlayed sidecar by anno with XXX_JAVA_OPTS, I will get two envs(XXX_JAVA_OPTS、JAVA_TOOL_OPTIONS) in final pod define. So I think default `JAVA_TOOL_OPTIONS` should be removed in swagent, this value default providered by [default config](https://github.com/apache/skywalking-swck/blob/master/operator/pkg/operator/manifests/injector/templates/annotations.yaml).

**About Why I overlay default envname to XXX_JAVA_OPTS**
When injector inject JAVA_TOOL_OPTIONS env, it will be picked up by every java process like jps, jmap.
Alough we can execute `export SW_AGENT_ENABLED=false` before these commands, out customers still think agent init output log is noisy, they prefer define XXX_JAVA_OPTS in their main java process start cmd.


wu-sheng:
@dashanji What is going on here?

dashanji:
> @dashanji What is going on here?

I'm going to merge this PR.

dashanji (approved):

PR #109, opened 2024-02-05
Bump up dependencies to fix CVE

kezhenxu94:


wu-sheng (approved):

wu-sheng:
@kezhenxu94 @dashanji Could you run a 0.9 release for swck? Or should we release 1.0? I think we have operator for a while.

dashanji:
> @kezhenxu94 @dashanji Could you run a 0.9 release for swck? Or should we release 1.0? I think we have operator for a while.

Okay, I will release 0.9.0 after the Spring Festival.

PR #113, opened 2024-04-13
Add CRD, Controller for SkyWalking Event Exporter

jichengzhi:
Related to [apache/skywalking#8937](https://github.com/apache/skywalking/issues/8937).

The controller (`EventExporterReconciler`) maintains a deployment and a configMap for each EventExporter CR under the hood. If filters and exporters (`EventExporter.spec.data`) are updated, the controller compares current config to the previous one to identify an update:

* prev config is empty and cur config is empty => no need to update
* prev config is empty and cur config is not empty => create a new configMap with cur config
* prev config is not empty and cur config is empty => delete the configMap
* prev config is not empty and cur config is not empty
  * `prev config != cur config` => re-create the configMap with cur config
  * `prev config == cur config` => no need to update

It is possible that the `EventExporter.Spec.Data` is empty if user ignores this field, in this case the event export will use the [default config](https://github.com/apache/skywalking-kubernetes-event-exporter/blob/9ad39c47eb347f13c0003d1ecd05c843c6d511ea/cmd/root.go#L50).

If config is changed, the `md5-data` label value in the pod template (`Deployment.spec.template.metadata`) is updated, which will trigger a deployment update after the rendered template is applied.


operator/apis/operator/v1alpha1/eventexporter_types.go
```diff
@@ -0,0 +1,78 @@
...
+	// +kubebuilder:validation:Required
+	Version string `json:"version,omitempty"`
+	// Image is the event exporter Docker image to deploy.
+	Image string `json:"image,omitempty"`
+	// Instances is the number of event exporter pods
+	// +kubebuilder:validation:Required
+	Instances int32 `json:"instances,omitempty"`
+	// Data of filters and exporters
+	// +kubebuilder:validation:Optional
+	Data string `json:"data,omitempty"`
```
fgksgf, line 37: How about renaming it to `Config`

operator/apis/operator/v1alpha1/eventexporter_types.go
```diff
@@ -0,0 +1,78 @@
...
+// EventExporterSpec defines the desired state of EventExporter
+type EventExporterSpec struct {
+	// Version of EventExporter.
+	// +kubebuilder:validation:Required
+	Version string `json:"version,omitempty"`
+	// Image is the event exporter Docker image to deploy.
+	Image string `json:"image,omitempty"`
+	// Instances is the number of event exporter pods
+	// +kubebuilder:validation:Required
+	Instances int32 `json:"instances,omitempty"`
```
fgksgf, line 34: `Replicas` would be better

operator/apis/operator/v1alpha1/eventexporter_webhook.go
```diff
@@ -0,0 +1,91 @@
...
+
+var _ webhook.Defaulter = &EventExporter{}
+
+// Default implements webhook.Defaulter so a webhook will be registered for the type
+func (r *EventExporter) Default() {
+	eventexporterlog.Info("default", "name", r.Name)
+
+	if r.Spec.Version == "" {
+		// use the latest version by default
+		r.Spec.Version = "latest"
```
fgksgf, line 50: Let's use a constant

operator/apis/operator/v1alpha1/eventexporter_webhook.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	if r.Spec.Image == "" {
+		r.Spec.Image = fmt.Sprintf("apache/skywalking-kubernetes-event-exporter:%s", r.Spec.Version)
```
fgksgf, line 54: Ditto.

operator/controllers/operator/eventexporter_controller.go
```diff
@@ -0,0 +1,210 @@
...
+	apiequal "k8s.io/apimachinery/pkg/api/equality"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/util/retry"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	runtimelog "sigs.k8s.io/controller-runtime/pkg/log"
+	"text/template"
```
fgksgf: sort the imports, otherwise lint will fail

operator/controllers/operator/eventexporter_controller.go
```diff
@@ -0,0 +1,210 @@
...
+// +kubebuilder:rbac:groups=operator.skywalking.apache.org,resources=eventexporters,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=operator.skywalking.apache.org,resources=eventexporters/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=operator.skywalking.apache.org,resources=eventexporters/finalizers,verbs=update
+// +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list
+// +kubebuilder:rbac:groups="",resources=events,verbs=create;patch
+
+func (r *EventExporterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := runtimelog.FromContext(ctx)
+	log.Info("=====================eventexporter reconcile started================================")
```
fgksgf, line 60: Let's also log the req.name and req.namespace

operator/controllers/operator/eventexporter_controller.go
```diff
@@ -0,0 +1,210 @@
...
+	if err := r.checkState(ctx, log, &eventExporter); err != nil {
+		l.Error(err, "failed to check sub resources state")
+		return ctrl.Result{}, err
+	}
+
+	return ctrl.Result{RequeueAfter: schedDuration}, nil
+
+}
+
+func (r *EventExporterReconciler) overlayData(ctx context.Context, log logr.Logger, eventExporter *operatorv1alpha1.EventExporter) (bool, error) {
```
fgksgf, line 102: Let's give the returned variables names, like `(changed bool, err error)` for better readability

operator/controllers/operator/eventexporter_controller.go
```diff
@@ -0,0 +1,210 @@
...
+		return errCol.Error()
+	})
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *EventExporterReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&operatorv1alpha1.EventExporter{}).
+		Owns(&apps.Deployment{}).
+		Owns(&core.ConfigMap{}).
```
fgksgf, line 210: I think if we do not add predicates, all deploy/configmaps change will trigger  eventExporter controller to reconcile. So we can remove them.

operator/apis/operator/v1alpha1/eventexporter_types.go
jichengzhi: Thx, updated!

operator/apis/operator/v1alpha1/eventexporter_types.go
jichengzhi: Thx, updated!

operator/apis/operator/v1alpha1/eventexporter_webhook.go
jichengzhi: Thx, updated!

operator/apis/operator/v1alpha1/eventexporter_webhook.go
jichengzhi: Thx, updated!

operator/controllers/operator/eventexporter_controller.go
jichengzhi: Thx, updated!

operator/controllers/operator/eventexporter_controller.go
jichengzhi: Thx, updated!

operator/controllers/operator/eventexporter_controller.go
jichengzhi: Thx, updated!

operator/controllers/operator/eventexporter_controller.go
jichengzhi: Thx, updated!

wu-sheng:
Should we consider adding e2e for this new feature?

jichengzhi:
Yes, I'm working on that, but still need some time to figure that out. Will commit e2e test ASAP

wu-sheng:
> Yes, I'm working on that, but still need some time to figure that out. Will commit e2e test ASAP

Don't worry. I just comment here as a reminder. There is nothing in urgent.

operator/apis/operator/v1alpha1/eventexporter_types.go
```diff
@@ -0,0 +1,78 @@
...
+	Replicas int32 `json:"replicas,omitempty"`
+	// Config of filters and exporters
+	// +kubebuilder:validation:Optional
+	Config string `json:"config,omitempty"`
+}
+
+// Important: Run "make" to regenerate code after modifying this file
+
+// EventExporterStatus defines the observed state of EventExporter
+type EventExporterStatus struct {
```
fgksgf, line 43: Maybe we need a field to indicate the name of the configmap that the event exporter is currently using. 

For different config content, the configmap will have different suffix like `event-exporter-cm-<hash>`. So every time the reconciler just need to hash the config in the spec, and compare it with the configmap name.

This has another advantage, when config is changed, we just need to calculate the new configmap name and update it in the deployment's spec, pods will be rolling updated automatically to use the new config.

operator/controllers/operator/eventexporter_controller.go
```diff
@@ -0,0 +1,210 @@
...
+					Name:       eventExporter.Name,
+					UID:        eventExporter.UID,
+				},
+			},
+			Labels: map[string]string{
+				"version":  eventExporter.Spec.Version,
+				"md5-data": newMd5,
+			},
+		},
+		Data: map[string]string{"config.yaml": eventExporter.Spec.Config},
```
fgksgf, line 146: If we create a new configmap for every new config in the spec, we can make the configmap immutable: https://kubernetes.io/docs/concepts/configuration/configmap/#configmap-immutable

operator/apis/operator/v1alpha1/eventexporter_types.go
jichengzhi: Thx for reply!  I wanted to exploit the default config in event exporter even though the [default address](https://github.com/apache/skywalking-kubernetes-event-exporter/blob/9ad39c47eb347f13c0003d1ecd05c843c6d511ea/assets/default-config.yaml#L42) won't work in k8s. Creating a configMap for all possible content is better because it facilitates config update handling (configMap always exist) and makes more sense if config is empty (no exporters >>> 100% failed exporter). I didn't figure this out before.

jichengzhi: Updated!

operator/controllers/operator/eventexporter_controller.go
```diff
@@ -0,0 +1,210 @@
...
+	Scheme   *runtime.Scheme
+	FileRepo kubernetes.Repo
+	Recorder record.EventRecorder
+}
+
+// +kubebuilder:rbac:groups=operator.skywalking.apache.org,resources=eventexporters,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=operator.skywalking.apache.org,resources=eventexporters/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=operator.skywalking.apache.org,resources=eventexporters/finalizers,verbs=update
+// +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list
```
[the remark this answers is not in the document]
jichengzhi: Thx, updated!

operator/controllers/operator/eventexporter_controller.go
jichengzhi: Thx, updated!

jichengzhi:
I added a simple e2e test case for `EventExporter` which tests whether the pod events (`Pulled`, `Created`, `Started`) of the demo service is reported to the `OAPServer`.

wu-sheng:
e2e needs to add into GHA control file, otherwise, it would not run in the PR process.

operator/controllers/operator/eventexporter_controller.go
```diff
@@ -0,0 +1,218 @@
...
+
+package operator
+
+import (
+	"context"
+	"fmt"
+	"text/template"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
```
fgksgf, line 26:
```suggestion
```

operator/controllers/operator/eventexporter_controller.go
```diff
@@ -0,0 +1,218 @@
...
+		if err := r.Status().Update(ctx, eventExporter); err != nil {
+			errCol.Collect(fmt.Errorf("failed to update status of EventExporter: %w", err))
+		}
+		return errCol.Error()
+	})
+}
+
+func configMapName(eventExporter *operatorv1alpha1.EventExporter) string {
+	return fmt.Sprintf("%s-eventexporter-cm-%s", eventExporter.Name, MD5Hash(eventExporter.Spec.Config))
+}
```
fgksgf, line 211: Is it possible that this exceeds the name length limit in k8s?

operator/controllers/operator/eventexporter_controller.go
```diff
@@ -0,0 +1,218 @@
...
+	err = r.Client.Get(ctx, client.ObjectKey{Namespace: eventExporter.Namespace, Name: oldConfigMapName}, &configmap)
+	if err != nil && !apierrors.IsNotFound(err) {
+		log.Error(err, "failed to get the eventexporter's configmap")
+		return true, err
+	}
+
+	if !apierrors.IsNotFound(err) {
+		if err = r.Client.Delete(ctx, &configmap); err != nil {
+			log.Error(err, "failed to delete eventexporter's configmap")
+			return true, err
```
fgksgf, line 125: We'd better to create the new CM first, then delete the old one. So the returned `changed: true` makes sense.

test/e2e/skywalking-components-with-eventexporter.yaml
```diff
@@ -0,0 +1,78 @@
...
+spec:
+  version: 9.5.0
+  instances: 1
+  image: apache/skywalking-ui:9.5.0
+  OAPServerAddress: http://skywalking-system-oap.skywalking-system:12800
+  service:
+    template:
+      type: ClusterIP
+    ingress:
+      host: demo.ui.skywalking
```
fgksgf, line 77: It seems we do not this in e2e?

fgksgf (commented):
Almost LGTM, just some nits, nice work! 

Do not forget to update the file `.github/workflows/go.yml` to make your added cases can be executed in CI.

fgksgf:
> ## RCQE-bot: Commit Quality Assessment 🔎
> I've carefully evaluated the recent commits, and here's the current state of the commit quality:
> 
> ### Metrics 📊
> * Previous Commits: diff hunk in file [operator/apis/operator/v1alpha1/eventexporter_types.go](https://github.com/apache/skywalking-swck/pull/113/files/d41f3c75f77368d7468ca3a8816c69ec59f44b55#diff-08e64e4e9e02c8371bf5173ecd944aa820a586af513d2c88a37680f117958919) in commit [ed272283fa32a6b960f8fbc9b5ce1a45f4ddf63d](https://github.com/apache/skywalking-swck/pull/113/commits/ed272283fa32a6b960f8fbc9b5ce1a45f4ddf63d)
> * Code Review: I think if we do not add predicates, all deploy/configmaps change will trigger eventExporter controller to reconcile. So we can remove them.
> 
> ### Suggestions 💡
> * Need Review Again 🚫 : The commits do not currently meet the quality standards. I recommend not merging them into the main branch until the above suggestions have been implemented and a subsequent review confirms the quality. Please address the points mentioned and submit the commits for another round of review.

There is indeed an issue with this commit. Thanks, will review again.

operator/controllers/operator/eventexporter_controller.go
jichengzhi: Thx for reply! 

The maximum length of the configMap name is [253 characters](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names), so it is possible to exceed the limit if length of the event exporter name is greater than 221 characters (253 - 32 for config MD5, without other suffix). 

Is it reasonable if I use `eventexporter-cm-{{ md5 }}` for all configMap names where `md5 = Md5Hash(eventExporter.Name + eventExporter.Spec.Config)`? EventExporter name is also hashed to make sure the generated name is unique.

test/e2e/skywalking-components-with-eventexporter.yaml
jichengzhi: Thx, updated!

operator/controllers/operator/eventexporter_controller.go
jichengzhi: Thx, updated!

operator/controllers/operator/eventexporter_controller.go
jichengzhi: Thx, updated!

jichengzhi:
> e2e needs to add into GHA control file, otherwise, it would not run in the PR process.

Thanks! It's fixed in the latest commit.

operator/controllers/operator/eventexporter_controller.go
fgksgf: In my opinion, it's better to have a fixed length suffix, just like the pods owned by k8s deployments. For the deployment `foo`, one of its pod might look like `foo-84ff48df47-lljb9`.

jichengzhi: Got it. But I think `-eventexporter-cm-<hash>` is a fixed length suffix since the MD5 hash is always 32 characters regardless of input length. Other hash functions generating fixed length hashes (SHA-1, SHA-256) generate longer outputs compared to MD5. We keep the hashed config as part of the suffix so that we don't need to compare the config content, which cannot be replaced by shorter random strings.

fgksgf: OK

fgksgf (approved):
LGTM, thanks for your contribution :)

jichengzhi:
Sorry, it happened so fast... In my last commit he configMap name is generated as `eventexporter-cm-<md5>` without event exporter name. Do I need to change it back to `<name>-eventexporter-cm-<md5>`?

jichengzhi:
> > Sorry, it happened so fast... In my last commit he configMap name is generated as `eventexporter-cm-<md5>` without event exporter name. Do I need to change it back to `<name>-eventexporter-cm-<md5>`?
> 
> It's ok to use `eventexporter-cm-<md5>`, since the configmap is namespaced resource.

Got it. Thank you for all your guidance!

dashanji:
Hi @jichengzhi, great job on the contribution! Could you please add a document detailing the CRD descriptions and instructions on how to use them under https://github.com/apache/skywalking-swck/tree/master/docs directory? This would be helpful to end users.

jichengzhi:
> Hi @jichengzhi, great job on the contribution! Could you please add a document detailing the CRD descriptions and instructions on how to use them under https://github.com/apache/skywalking-swck/tree/master/docs directory? This would be helpful to end users.

Sure! I'll finish the doc before Apr 24th.

dashanji:
Thanks！

PR #116, opened 2024-04-22
Add doc for EventExporter

jichengzhi:
Add doc explaining how to use the EventExporter CRD added in #113.

### DIFF
--- a/docs/examples/event-exporter.md
+++ b/docs/examples/event-exporter.md
@@ -172,7 +172,7 @@ DEBUG done: rendered event is: uuid:"8d8c2bd1-1812-4b0c-8237-560688366280" sourc
 
 | name     | description                                            | default value                                        |
 |----------|--------------------------------------------------------|------------------------------------------------------|
-| image    | Docker image of the event exporter.                    | `apache/skywalking-kubernetes-event-exporter:latest` |                                                                                                           | `"skywalking-system-oap.skywalking-system:11800"`                       |
+| image    | Docker image of the event exporter.                    | `apache/skywalking-kubernetes-event-exporter:latest` |                                                                                                          
 | replicas | Number of event exporter pods.                         | `1`                                                  |
 | config   | Configuration of filters and exporters in YAML format. | `""`                                                 |
 
@@ -227,7 +227,7 @@ exporters:
 
 | name      | description                                                                                                                         | example          |
 |-----------|-------------------------------------------------------------------------------------------------------------------------------------|------------------|
-| reason    | Filter events of the specified reason, regular expression like `"Killing\|Killed"` is supported.                                    | `""`             |                                                                                                           | `"skywalking-system-oap.skywalking-system:11800"`                       |
+| reason    | Filter events of the specified reason, regular expression like `"Killing\|Killed"` is supported.                                    | `""`             |                                                                                                       
 | message   | Filter events of the specified message, regular expression like `"Pulling container.*"` is supported.                               | `""`             |
 | minCount  | Filter events whose count is >= the specified value.                                                                                | `1`              |
 | type      | Filter events of the specified type, regular expression like `"Normal\|Error"` is supported.                                        | `""`             |
--- a/docs/examples/event-exporter.md
+++ b/docs/examples/event-exporter.md
@@ -1,0 +1,271 @@
+# SkyWalking Kubernetes Event Exporter User Guide
+
+[SkyWalking Kubernetes Event Exporter](https://github.com/apache/skywalking-kubernetes-event-exporter) is able to watch,
+filter, and send Kubernetes events into the Apache SkyWalking backend.
+
+## Demo
+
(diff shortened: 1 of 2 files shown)


docs/examples/event-exporter.md
```diff
@@ -0,0 +1,271 @@
...
+        endpoint: ""
+      message: "{{ .Event.Message }}"
+    address: "127.0.0.1:11800" 
+```
+
+### Filter Config
+
+| name      | description                                                                                                                         | example          |
+|-----------|-------------------------------------------------------------------------------------------------------------------------------------|------------------|
+| reason    | Filter events of the specified reason, regular expression like `"Killing\|Killed"` is supported.                                    | `""`             |.                                                                                                           | `"127.0.0.1:11800"`                       |
```
dashanji, line 230: What's the meaning of `"127.0.0.1:11800"`?

jichengzhi: The TCP port of exposed by the oap service. I used the [default value](https://github.com/apache/skywalking-kubernetes-event-exporter/blob/9ad39c47eb347f13c0003d1ecd05c843c6d511ea/assets/default-config.yaml#L42), maybe I need to change it to a more reasonable address like `skywalking-system-oap.skywalking-system:11800`

jichengzhi: I didn't include a link to the [default config file](https://github.com/apache/skywalking-kubernetes-event-exporter/blob/main/assets/default-config.yaml#L42) because if you ignore the `EventExporter.Spec.Config` field, an empty string is used as configuration instead of the default config mentioned in skywalking-kubernetes-event-exporter. So there is no "default config" in EventExporter, the default is `""`. That's also why I used "example" for column name, not "default value".

docs/examples/event-exporter.md
```diff
@@ -0,0 +1,271 @@
...
+        endpoint: ""
+      message: "{{ .Event.Message }}"
+    address: "skywalking-system-oap.skywalking-system:11800" 
+```
+
+### Filter Config
+
+| name      | description                                                                                                                         | example          |
+|-----------|-------------------------------------------------------------------------------------------------------------------------------------|------------------|
+| reason    | Filter events of the specified reason, regular expression like `"Killing\|Killed"` is supported.                                    | `""`             |.                                                                                                           | `"skywalking-system-oap.skywalking-system:11800"`                       |
```
dashanji, line 230:
```suggestion
| reason    | Filter events of the specified reason, regular expression like `"Killing\|Killed"` is supported.                                    | `""`             |
```

The extra field `"skywalking-system-oap.skywalking-system:11800"` will not be rendered by markdown, could you please check it?

docs/examples/event-exporter.md
```diff
@@ -0,0 +1,271 @@
...
+kubectl logs -f skywalking-system-eventexporter-566db46fb6-npx8v -n skywalking-system
+...
+DEBUG done: rendered event is: uuid:"8d8c2bd1-1812-4b0c-8237-560688366280" source:{service:"skywalking-system-oap" serviceInstance:"skywalking-system-oap-68bd877f57-zs8hw"} name:"Started" message:"Started container oap" startTime:1713795214000 endTime:1713795214000 layer:"K8S"
+```
+
+## Spec
+
+| name     | description                                            | default value                                        |
+|----------|--------------------------------------------------------|------------------------------------------------------|
+| image    | Docker image of the event exporter.                    | `apache/skywalking-kubernetes-event-exporter:latest` |.                                                                                                           | `"skywalking-system-oap.skywalking-system:11800"`                       |
```
dashanji, line 175: ditto

docs/examples/event-exporter.md
jichengzhi: Thx, updated!

docs/examples/event-exporter.md
jichengzhi: Thx, updated!

docs/examples/event-exporter.md
dashanji: Hi @jichengzhi, it looks like the field `"skywalking-system-oap.skywalking-system:11800"` still exists, could you please delete it?

jichengzhi: Sorry I missed that before, should be removed now.

dashanji (approved):
LGTM

wu-sheng:
@dashanji How is the e2e here? 

> Continuous Integration / e2e tests(oap+ui+agent+internel-storage) (pull_request) 

This fails, and this PR seems not related.

dashanji:
> @dashanji How is the e2e here?

I'm not sure if recent e2e tests were caused by k8s or skywalking components.  Maybe need a further exploration.

wu-sheng:
Let's fix the SWCK in helm first. Then we could address this.

dashanji:
> Let's fix the SWCK in helm first. Then we could address this.

Okay.

PR #120, opened 2024-12-23
ci: add dependabot configuration

kezhenxu94:
We can re-generate the license file but pushing the changes to the PR that dependabot opened won’t trigger the CI checks, this is by design in GitHub Actions to avoid infinite recursive workflow runs, so in this PR, I just assign myself and after dependabot open the PR, I’ll take the remaining steps to update the license and merge it.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -22,9 +22,17 @@ updates:
       interval: "daily"
     assignees:
       - "kezhenxu94"
+    groups:
+      actions-deps:
+        patterns:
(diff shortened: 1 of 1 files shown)


wu-sheng (approved):

kezhenxu94:
Converted to draft as I want to add groups to dependabot, so that multiple dependencies can be opened in a single PR, reducing maintenance workload.

wu-sheng (approved):

PR #140, opened 2025-02-21
Replace deprecated APIs with recommended ones

kezhenxu94:


wu-sheng (approved):
According to @hanahmily BanyanDB operator doesn't provide unique feature as an operator from helm.
And we have to maintain it. We had a discussion, and it seems that we could remove the whole banyandb operator and docs. 
WDYT?

kezhenxu94:
> According to @hanahmily BanyanDB operator doesn't provide unique feature as an operator from helm. And we have to maintain it. We had a discussion, and it seems that we could remove the whole banyandb operator and docs. WDYT?

I’ll defer this to @hanahmily or @dashanji, this PR to make it smooth for the dependabot to upgrade dependencies, to unblock #138 .

PR #214, opened 2026-08-30
Release the Helm chart from this repository, and move the stack to SkyWalking 11 + Horizon

wu-sheng:
Brings the SWCK Helm chart into this repo, automates the release, and moves the whole stack
onto SkyWalking 11 with the Horizon UI.

The chart has lived in `apache/skywalking-helm` and drifted from the operator it installs:
the RBAC and webhook manifests were hand-maintained copies of `operator/config`, so every
API change had to be mirrored by hand in another repository and usually wasn't. Here the
generated pieces are generated, and CI fails when they drift.

### The chart

`chart/skywalking-swck` installs the operator and, optionally, the custom metrics adapter.
`crds/`, `operator-manager-role.yaml` and `operator-webhook.yaml` are produced by
`tools/generate-chart.sh` from `operator/config` via kustomize; `make chart-check`
regenerates into a scratch directory and diffs, so a CRD or RBAC change that skips the chart
fails CI rather than shipping a chart that under-permits its own operator.

Certificates are plain resources rather than Helm hooks — as hooks they deadlock
`helm install --wait`, because the webhook waits for a certificate whose hook cannot run
until the install completes. `kubeVersion` is `>=1.21.0-0`, which is what the CRDs require.

### Releasing — and a workflow that had never run

`tools/releasing/release.sh` and `release-passed.sh` follow `apache/skywalking`: the first
builds and signs the vote candidate, the second promotes it after the vote passes.

While reviewing this, the publish workflow turned out never to have run at all. Its three
`docker/*` actions were pinned to commit SHAs absent from the ASF allow-list, and the
gateway rejects such a workflow *before any job starts* — so it records `startup_failure`, a
run with zero jobs rather than a failing step, which is easy to read as "nothing to do".
Every run on master since **2026-06-16** is in that state. Repinned to approved SHAs, with
the constraint recorded next to them so a future bump doesn't silently re-break it.

Docker publication triggers on a *published* GitHub release rather than a tag, gated on
`gh release view`. Chart publication to GHCR is independent of Docker Hub on the snapshot
path, but now waits for `publish-combined-image` on the release path, because a release
chart keeps its Docker Hub defaults and would otherwise advertise an image that did not
exist yet.

Chart and image share a Docker Hub namespace, so charts carry a `-helm` suffix; still valid
SemVer.

### arm64 was never arm64

`build/images/Dockerfile.release` hardcoded `-linux-amd64` regardless of the buildx
platform. `apache/skywalking-swck:0.10.0` advertised both architectures, but the arm64
manifest reused byte-identical amd64 layer digests and the `/manager` inside it was
`ELF 64-bit LSB executable, x86-64` — an arm64 node got `exec format error`. It now uses
`TARGETARCH`, and CI extracts the binary from each published manifest and asserts its ELF
machine type, so a repeat is a build failure rather than a discovery in production.

### Storage is mandatory, because there is no fallback

SkyWalking removed H2 **permanently in 10.2.0**, not in 11.0.0. With this operator's floor at
10.4.0, no supported OAP has an embedded storage and there is nothing to fall back to: an
`OAPServer` with nowhere to write starts, dials a BanyanDB on `127.0.0.1:17912` and never
becomes ready.

That is now refused at admission rather than described in a comment. Setting `SW_STORAGE`
directly in `spec.config` still counts as having chosen a storage. **Breaking**: an
`OAPServer` with neither is rejected — it could never have worked. Turning it into a hard
failure immediately exposed three samples (including `minimal.yaml`, still pinned to 9.1.0,
below the supported floor), two doc examples and one e2e case in exactly that state, all of
which had been shipping as things a user could apply and watch never become ready.

The operator also gains BanyanDB as a first-class `Storage` type —
`SW_STORAGE_BANYANDB_TARGETS`, plus credentials from a secret — and warns, rather than
rejects, below OAP 10.4.0.

### Horizon

The UI stack is Horizon only: the Booster branches are gone and `kind` is an enum of one.

The generated `horizon.yaml` now follows [Horizon's own shipped reference
config](https://github.com/apache/skywalking-horizon-ui/blob/main/horizon.yaml), which keeps
`roles`, `landingByRole` and `performance` at `:null` — "fall through to the built-in
default" — rather than enumerating them. Ours enumerated them, and the copy had drifted from
what it was copying: `viewer` was granted 6 of the 12 permissions Horizon gives that role,
and the admin landing route was `/admin/cluster`, which Horizon has no route for (its cluster
page is `/operate/cluster`), so signing in as admin landed on "No route matches".

Reading that reference also surfaced something missing. `templates.mode` selects whether
Horizon reads dashboards from OAP's store or from the bundle in its image, and **OAP 10.x
requires `readonly`** — OAP 10 manages templates over legacy query-port GraphQL and Horizon
implements only OAP 11's REST protocol, so in `live` mode the store is unreachable and
Horizon blocks every layer-driven page, Traces most visibly. We emitted no `templates.mode`
at all, so a UI deployed against a supported OAP came up with its layer pages blocked. New
`UI.spec.templatesMode` carries it; the operator cannot infer it, since a `UI` holds an OAP
address, not an OAP version.

`OAPServerAdminAddress` and `OAPServerZipkinAddress` are no longer guessed — on 10.x the port
they were derived from belongs to a different service.

### Operator fixes

`OAPServerConfig.OverlayEnv` assigned its own env over the container's, discarding the
storage wiring the OAPServer controller had injected. Fixing that by overlaying introduced a
second bug — the overlay could add and change but never *remove*, so a name dropped from
`spec.env` lingered while the md5 label reported convergence. It now records what it
displaced and restores it, including template values like `JAVA_OPTS=-Xmx2048M`.

The OAP startup probe allowed 110s, which is not enough for an OAP 11 schema init.

### e2e

Twelve cases, listed in `docs/en/guides/e2e.md`, covering deployment, CRDs, sidecar agent
injection, HPA against both OAP and satellite metrics, the event exporter, dynamic
configuration, both storages, and a chart install/upgrade/uninstall lifecycle. Ten install
the operator through the chart; `oap-ui-agent` stays on kustomize because `operator/config`
still ships in the binary release. Every image is pinned in `test/e2e/env` and substituted
with `envsubst`. `banyandb-e2e-tests` now gates the required check — it never did, which
mattered little when H2 existed and matters a lot now.

The satellite stays pinned to `1987e1d5` rather than the newer `35bfaff6`: on that build
`sw_stl_grpc_server_connection_count` stops tracking live connections — five agents attached,
the satellite visibly serving all of them, the metric peaking at 3 and decaying to 1 — so the
HPA reading it never scaled. Reasoning recorded in `test/e2e/env`.

### Layout and docs

`hack/` is split by audience: `tools/` for what developers run, `test/tools/` for what the
tests run. `docs/` is restructured to match `apache/skywalking` — `docs/en/{concepts-and-designs,
setup,examples,guides,changes}` with a `menu.yml` the website renders — and `CHANGES.md`
becomes `docs/en/changes/`.

### Known, and deliberately left out

The metrics adapter discards the request context (`_ context.Context`) and queries OAP with
`p.ctx`, a urfave/cli `*cli.Context` rather than a `context.Context`, with no HTTP client
timeout — so a hung OAP query has nothing to cancel it. It also reads `p.metricDefines`
without the lock `ListAllExternalMetrics` takes for the same field, and
`--metric-filter-regex` defaults to empty so it advertises all 1754 OAP 11 metrics. All three
predate this PR and belong in their own issue.


kezhenxu94 (commented):
The files under `crds` folder are previously generated from this repo and copied to the skywalking-helm, now that we move the Helm Chart here, we might consider generate the `crds` folder on the flight when building the Helm Chart, this can avoid disconnecting from the codes and the Helm Chart

kezhenxu94:
> The files under `crds` folder are previously generated from this repo and copied to the skywalking-helm, now that we move the Helm Chart here, we might consider generate the `crds` folder on the flight when building the Helm Chart, this can avoid disconnecting from the codes and the Helm Chart

Or at least, we need a CI check to make sure the generated CRDs reflect what the codes generate

test/e2e/skywalking-components-with-eventexporter.yaml
```diff
@@ -23,7 +23,7 @@ metadata:
 spec:
   version: 9.5.0
   instances: 1
-  image: apache/skywalking-oap-server:9.5.0
+  image: OAP_IMAGE
```
kezhenxu94: I’d rather not to use such bare string as "placeholder", something like `${OAP_IMAGE}`/`$OAP_IMAGE` would be more intuitive, and we don’t even need to do the manual `sed` command in `render.sh`, just use something like `source ./test/e2e/env && envsubst < test/e2e/skywalking-components-with-eventexporter.yaml | kubectl apply -f -`;

the idea is, using something well known to the world (form of `${PLACEHOLDER}`, instead of `PLACEHOLDER`, and tools like `envsubst` exactly for this purpose), instead of reinventing something that needs maintainers to learn (the render.sh script)

wu-sheng:
Thanks — both addressed.

**`${VAR}` + `envsubst` instead of `render.sh`.** Agreed, and `render.sh` is deleted. Call sites are now plain:

```shell
envsubst "$E2E_IMAGES" < test/e2e/skywalking-components.yaml | kubectl apply -f -
```

`test/e2e/env` is loaded into every step by infra-e2e's `setup.init-system-environment`, so nothing needs sourcing. One wrinkle worth knowing: `E2E_IMAGES` (defined at the bottom of `env`) is the explicit list of image variables, and passing it restricts substitution to those names. A bare `envsubst` would also eat things like `namespace: "^skywalking-system$"` in the EventExporter filter.

**Generated `crds/`.** Both, already — generated *and* checked. `hack/generate-chart.sh` produces `crds/`, the manager `ClusterRole` and both webhook configurations from the operator sources, and `make chart-check` regenerates into a scratch directory and diffs, so a hand-edit is a build failure. It runs in the new **Helm chart** CI job on every PR, and again in the publish workflow before anything is pushed.

The inputs go through `kustomize build` rather than being read straight off `operator/config`, because those kustomizations carry patches that change the result — `config/rbac` prepends a rule via JSON 6902, and the strategic merge in `ns_selector_patch.yaml` is also what collapses the duplicate `meventexporter.kb.io` entry controller-gen emits, which the API server rejects. Copying the raw files would have shipped that duplicate.

It already caught real drift: the charts in skywalking-helm were missing the `eventexporter` webhook entirely.

kezhenxu94 (dismissed):

kezhenxu94 (approved):

